PR #59, opened 2020-04-10
Aymeric

AymericNoel:
Tests completed.


src/Trakx.Tests/Data/MockCreator.cs
```diff
@@ -118,5 +119,19 @@ public TransactionReceipt GetTransactionReceipt()
             };
             return transactionReceipt;
         }
+
+        public IIndexValuation Get_index_valuation(IIndexComposition composition, string quoteCurrency = "usdc")
```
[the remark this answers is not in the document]
monsieurleberre: OK maybe the bot is right :) this class only has non static method and it could be changed to static...
However for non Test method names, can you please use Pascal casing? GetIndexValuation?

FYI the tests tend to have very long names so in that case my_method_should_do_this is a bit more justified. Thanks.

src/Trakx.MarketData.Server/Models/IndexPricedModel.cs
```diff
@@ -30,7 +30,7 @@ public partial class IndexPricedModel
 
             var indexPriced = new IndexPricedModel()
             {
-                Symbol = initialValuation.IndexComposition.IndexDefinition.Symbol.ToUpper(),
+                Symbol = initialValuation.IndexComposition.IndexDefinition.Symbol.ToLower(),
```
monsieurleberre, line 33: Not sure we should change the code, I think the code on the exchange is case sensitive unfortunately

src/Trakx.MarketData.Server/Controllers/IndexDataController.cs
```diff
@@ -48,7 +48,7 @@ public async Task<ActionResult<string>> IndexDetailsPriced([FromQuery] string in
             var composition = await _indexProvider.GetCurrentComposition(indexSymbol);
             
             if (composition == default)
-                return $"failed to retrieve details for index {indexSymbol}";
+                return new JsonResult($"failed to retrieve details for index {indexSymbol}");
```
monsieurleberre, line 51: 👍

src/Trakx.MarketData.Server/Controllers/IndexDataController.cs
```diff
@@ -70,7 +70,7 @@ public async Task<ActionResult<string>> IndexDetailsPriced([FromQuery] string in
 
             var indexPriced = IndexPricedModel.FromIndexValuations(issuanceValuation, currentValuation);
             indexPriced.ComponentDefinitions.ForEach(d => d.IconUrl = iconBySymbol[d.Symbol.ToLower()]);
-
+            indexPriced.ComponentDefinitions.ForEach(d => d.Symbol = d.Symbol.ToLower());
```
monsieurleberre, line 73: are you sure? is this just to change the case? maybe we can do it, but then it should happen when you convert the `IComponentDefinition` into a `ComponentModel` in componentModel.cs

AymericNoel: I tested without this line and the test on the recognition of the symbol of the component failed.

```
finalResults.ComponentDefinitions[0].Symbol.Should()
                .Be(composition.ComponentQuantities[0].ComponentDefinition.Symbol);
```

monsieurleberre: OK is it working if you don't do case sensitive comparison?

AymericNoel: Yes it is.

PR #99, opened 2020-05-07
Wrapping history razor

jmc171144:
Very small PR about the razor page "WrappingHistory". 
There's no point in my opinion to make tests for the moment, because it's only simulating the display of the data of the db (for the moment).


src/Trakx.IndiceManager.Client/Pages/Wrapping/WrappingHistory.razor
```diff
@@ -1,7 +1,92 @@
...
+    }
+</table>
+
+<MatDialog @bind-IsOpen="@dialogIsOpen">
+    <MatDialogActions>
+        <MatButton OnClick="@(e => { dialogIsOpen = false; id = 0; })" Icon="close"></MatButton>
+    </MatDialogActions>
+    <MatDialogTitle>Transaction details</MatDialogTitle>
+    <MatDialogContent>
+        <p>Transaction Id on @WrappingHistoryList[id].FromCurrency Chain: @TxIdFromCurrency</p>
```
monsieurleberre, line 34: Maybe one of them should be saying `Ethereum Chain` rather than `wXRP Chain` ?

PR #100, opened 2020-05-07
dynamic indice history

jmc171144:
Same here but with Indices.
_Very small PR about the razor page "WrappingHistory".
There's no point in my opinion to make tests for the moment, because it's only simulating the display of the data of the db (for the moment)._


src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
```diff
@@ -2,6 +2,124 @@
...
+<MatDialog @bind-IsOpen="@dialogIsOpen">
+    <MatDialogActions>
+        <MatButton OnClick="@(e => { dialogIsOpen = false; onClickId = 0; })" Icon="close"></MatButton>
+    </MatDialogActions>
+    <MatDialogTitle>Summary of the @indicelList[onClickId].Name Indice</MatDialogTitle>
+    <MatDialogContent>
+        <p><strong>Name: </strong>@indicelList[onClickId].Name</p>
+        <p><strong>Symbol: </strong>@indicelList[onClickId].Symbol</p>
+        <p><strong>Id: </strong>@indicelList[onClickId].Id</p>
+        <p><strong>Nav: </strong>@indicelList[onClickId].Nav</p>
```
monsieurleberre, line 35: Cool, this NAV is the NAV at creation or the current one? Or do we want to put both?

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
```diff
@@ -2,6 +2,124 @@
...
+        <p><strong>Indice Address: </strong>"0x00000000000000000"</p>
+        <p><strong>Creation Date: </strong>@indicelList[onClickId].Date</p>
+        <p><strong>Composition: </strong></p>
+        <ol>
+            @foreach (var e in indicelList[onClickId].ComponentList)
+            {
+                <li>@e.Symbol: @e.Weight (USDc)</li>
```
monsieurleberre, line 42: Here it is called weight, but I think you show the component's value per unit of index.
Maybe we should put both, and if we put values and weight, maybe we can do `Value At Creation` and `Creation Value`
No big deal for now 👍

PR #101, opened 2020-05-07
Adding Indice Supply controller

AymericNoel:
There is not many routes but for the moment, I think that is enough.


src/Trakx.IndiceManager.Server/Controllers/IndiceSupplyController.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <returns>An object with a response 201 if the adding was successful</returns>
+        [HttpPost]
+        public async Task<ActionResult<bool>> SaveTransaction([FromBody] IndiceSupplyTransactionModel transactionToSave)
+        {
+            if (string.IsNullOrWhiteSpace(transactionToSave.User))
+                return BadRequest("There is no user attached to this transaction. Please try again.");
+
+            if (transactionToSave.IndiceComposition == null || transactionToSave.TransactionType == null ||
+                transactionToSave.IndiceQuantity == null)
+                return BadRequest("Some mandatory parameters are missing, please try again.");
```
monsieurleberre, line 35: No big deal, but that could be validation logic on the side of the IndiceSupplyTransactionModel. Then if you need to validate it somewhere else, you can reuse that :)

AymericNoel: hum, I see, I will modify that ! Thx

PR #107, opened 2020-05-14
Finishing manager + tests + refactoring

AymericNoel:
Short PR to finishing manager to issue and redeem indices.


AymericNoel:
I've tried to simplify the calls for the services like we said few days ago by regrouping them, but that failed...

src/Trakx.IndiceManager.Server/Models/IndiceSupplyTransactionModel.cs
```diff
@@ -39,12 +43,28 @@ public IndiceSupplyTransactionModel(IIndiceSupplyTransaction transaction)
 
         public bool IsValid()
         {
-            return IndiceComposition != default 
+            return IndiceComposition != null 
                    && !string.IsNullOrWhiteSpace(User)
                    && !string.IsNullOrWhiteSpace(SenderAddress)
-                   && CreationTimestamp != default
-                   && TransactionType != default 
+                   && CreationTimestamp != null
```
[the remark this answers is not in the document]
monsieurleberre: he's right, DateTime, like SupplyTransactionType are never `null` only classes can be `null`.
so `CreationTimestamp != default` means `CreationTimestamp != DateTime.MinDate` 
whereas `CreationTimestamp != null` means `true`
then he should have been alarmed at the next line
`SupplyTransactionType` is an enum and can't be null, `SupplyTransactionType?` is a nullable enum and can be null.
So to avoid this kind of confusion, when you really mean that you don't want the value to be its default, please use `default` instead of `null`. My opinion is that you should avoid using `null` in the code, `default` is often what you really mean, and if not, you need to ask yourself what you are really trying to do.

src/Trakx.IndiceManager.Server/Models/IndiceSupplyTransactionModel.cs
```diff
@@ -39,12 +43,28 @@ public IndiceSupplyTransactionModel(IIndiceSupplyTransaction transaction)
 
         public bool IsValid()
         {
-            return IndiceComposition != default 
+            return IndiceComposition != null
```
monsieurleberre, line 46: Can you please try to use `default` when trying to express that a variable has not been set?
Here there are 2 issues because of that
1) SupplyTransactionType ~?~ is not nullalbe, so it cannot be null, it is not a class type, so it will default to a value that is valid and it can be hard to notice. If you say it is nullabe using `SupplyTransactionType?` then it will default to null, and can break the code if it was not set intentionally, which is better than running the code with false assumption.
2) Same for CreationTimeStamp, it cannot be null, it will be defaulted to 01/01/01. I think here we could be happy that the transaction will be considered by the rest of the code with such a date, but thinking about it, it is probably best to make it nullable too.

Then because we can be changing the code between `struct` and `class` in some other area, wherever we would check for null would need to change. That is why it is best to check for default in my opinion.

jmc171144: _Trakx-bot like this_ 👍

PR #109, opened 2020-05-18
Syncfusion

monsieurleberre:
OK I have not done the Wrapping History page, but I think it is mainly a placeholder for now.
Also, as you can admire, my css skills are 💩💩💩💩💩💩💩 
Sorry about that 🙏 
Apart from that, I think it is quite an improvement, hopefully the little bits of testing I've done are clear and can be used as templates  for the future pages.


src/Trakx.IndiceManager.Client.Tests/Wrapping/WrapUnwrapTokenTests.cs
```diff
@@ -1,200 +1,183 @@
...
+            (await Dispatch(() => Component.Instance.EditContext.Validate()))
+                .Should().BeFalse();
+
+            _model.NativeAddress = "valid";
+            (await Dispatch(() => Component.Instance.EditContext.Validate()))
+                .Should().BeFalse();
+
+            _model.EthereumAddress = _mockCreator.GetRandomAddressEthereum();
+            (await Dispatch(() => Component.Instance.EditContext.Validate()))
+                .Should().BeFalse();
```
jmc171144, line 163: Do you know why when the model is correct (adding an initialization for ToCurrency) the .Should().BeTrue() doesn't work ?

monsieurleberre: Oh shit, I wanted to end up on a True here ... you are right I forgot one step
I think it stays false because Amount is out of range
```csharp
[Required, Range(1e-4, 10)]
public decimal? Amount { get; set; }
```

monsieurleberre: OK I have fixed it :)

jmc171144: oh yes I think too, it should works with True with a lower amount

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -1,163 +1,162 @@
...
-
-    public IndiceCreationModel myModel = new IndiceCreationModel();
-    public IndiceCreationModel.ComponentModel NewComponent = new IndiceCreationModel.ComponentModel();
-    private bool atLeastOneComponent = true;
-
-    [Parameter] public EventCallback<MouseEventArgs> OnCalculateNavSubmitClick { get; set; }
-
-    private Task HandleOnSubmitClick(MouseEventArgs args)
+#nullable disable
+    public SfTextBox NameTextBox { get; set; }
```
jmc171144, line 99: Syncfusion allows to create special types directly linked to their Html components ?

src/Trakx.IndiceManager.Client.Tests/Wrapping/WrapUnwrapTokenTests.cs
monsieurleberre: It is a bit simplistic and we have to be able to change the max and min amount based on the currency later. Maybe this can be a way to look at it:
https://docs.fluentvalidation.net/en/latest/async.html

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
monsieurleberre: so it is actually true for all Blazor components, you declare `@ref="ComponentAsProperty"` on its attribute and then you can access a typed `ComponentAsProperty` everywhere, including in tests 🍾

jmc171144: Nice !

PR #112, opened 2020-05-20
Adding coinbase library

AymericNoel:


src/Coinbase.Custody.Package/ExtensionsForFlurl.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using Flurl;
+using Flurl.Http;
+
+namespace Coinbase.Custody.Package
+{
+    public static class ExtensionsForUrl
```
monsieurleberre, line 7: Convention in the rest of the code is call these classes FlurlExtensions.
Not sure these methods add any value either (maybe the last one does)

src/Coinbase.Custody.Package/HeaderNames.cs
```diff
@@ -0,0 +1,36 @@
...
+    public class PaginationOptions
+    {
+        public int? Limit { get; set; }
+        public string Order { get; set; }
+        public string StartingAfter { get; set; }
+        public string EndingBefore { get; set; }
+    }
+
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public static class ExtensionsForFlurlUrl
```
monsieurleberre, line 22: Another class with _almost_ the same name 🤔🤔

src/Coinbase.Custody.Package/Program.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Threading.Tasks;
+using Coinbase.Custody.Package.Models;
+using Flurl.Http;
+using Trakx.Tests.Tools;
+
+namespace Coinbase.Custody.Package
+{
+    class Program
+    {
+        static async Task Main(string[] args)
```
monsieurleberre, line 12: See previous comments, you are writing a library, and a good way to see how it should be integrated in a program is to write an integration test.

src/Coinbase.Custody.Package/TimeHelper.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace Coinbase.Custody.Package
+{
+    public static class TimeHelper
```
monsieurleberre, line 5: Here you should have done an extension method 😋😋 sorry to be a pain in the 🍑 😅

src/Coinbase.Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM mcr.microsoft.com/dotnet/core/aspnet:3.1-buster-slim AS base
```
monsieurleberre, line 1: This file doesn't make  any sense to me, you are writing a library, not an executable. Sorry if I miss something.

.github/workflows/dockerpublish.coinbase.yml
```diff
@@ -0,0 +1,65 @@
+name: Docker Build and Publish Coinbase Custody Package
```
monsieurleberre, line 1: This file doesn't make sense to me, why would you want to dockerise a library?

monsieurleberre:
We will do this step by step starting with #113

PR #115, opened 2020-05-26
Adding coinbase Authentification

AymericNoel:
I have added a simple system to be authentify.

I will add test soon, but I don't know how yet !


src/Trakx.Coinbase.Custody.Client/ApiKeyConfig.cs
```diff
@@ -0,0 +1,44 @@
...
+
+namespace Trakx.Coinbase.Custody.Client
+{
+    public class ApiKeyConfig : IApiKeyConfig
+    {
+        private readonly string _apiKey;
+        private readonly string _apiPassphrase;
+        public void EnsureValid()
+        {
+            if (string.IsNullOrWhiteSpace(_apiKey)) throw new ArgumentNullException(nameof(_apiKey), "The API Key must be specified.");
```
monsieurleberre, line 15: You can reuse the logic from the later `ServiceConfiguration` file with Ardalis Guards instead.
Also not mega important as we are better safe than sorry, but are you doing the same check several times?

AymericNoel: Yes, it's redundant, I will correct that

src/Trakx.Coinbase.Custody.Client.Tests/Unit/ApiKeyConfigTest.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        private readonly HttpTest _httpTest;
+
+        public ApiKeyConfigTest()
+        {
+            _httpTest = new HttpTest();
+        }
+        [Fact]
+        public void ApiKeyConfig_should_return_error_if_apiKey_passphrase_are_empty()
+        {
```
monsieurleberre, line 23: You should create an action and use ShouldThrow () instead :)

PR #117, opened 2020-05-28
From table to datagrid

jmc171144:
Very small PR that replaces the data grid tables, unfortunately the routes don't exist yet so I couldn't do much... But we're going to fix it
I prefer to do a small PR and move on to something else so as not to mix up


src/Trakx.IndiceManager.Client/Pages/Wrapping/WrappingBalance.razor
```diff
@@ -54,4 +42,17 @@
...
 
+    public List<WrappingBalanceModel> GridData { get; set; }
+
+    #region Overrides of ComponentBase
+
+    /// <inheritdoc />
+    protected override void OnInitialized()
+    {
+        base.OnInitialized();
+        GridData = WrappingBalanceList;
```
monsieurleberre, line 53: You should delete this GridData as you bind the data grid to the `WrappingBalanceList `. Currently it creates a bit of confusion.

src/Trakx.IndiceManager.Client/Pages/Wrapping/WrappingHistory.razor
```diff
@@ -1,44 +1,45 @@
...
-        </tr>
-    }
-</table>
+<SfGrid DataSource="@WrappingHistoryList" Toolbar="@(new List<string>() { "Add", "Edit", "Delete", "Update", "Cancel", "Search" })" AllowPaging="true">
+    <GridEditSettings AllowAdding="true" AllowDeleting="true" AllowEditing="true" />
+    <GridTemplates>
+        <DetailTemplate>
+            @{
+                var wrappingTx = (context as WrappingHistoryModel);
+                <h3>Wrapping Transaction @wrappingTx.Id details</h3>
```
monsieurleberre, line 12: Maybe that one too deserves to be a DataGrid, rather than an html table 🤔 
But to be honest, I am not sure there is a point having a single row inside each row right? 
More columns would be better, especially when you are exporting the data to excel and then doing some processing on it.
Am I missing something?

src/Trakx.IndiceManager.Client/Pages/Wrapping/WrappingHistory.razor
```diff
@@ -56,12 +57,12 @@
     void OpenDialog(string _id)
```
monsieurleberre, line 57: I think that should be cleaned too if it is not used?

PR #120, opened 2020-05-29
Add cryptocompare prices from REST

monsieurleberre:
Add crypto compare as source of price before Messari => 
1) cached price from CryptoCompare websocket
2) price from CryptoCompare REST
3) Messari
4) CoinGecko
Also started a Common.Tests project


src/Trakx.Common.Tests/Unit/Pricing/NavCalculatorTests.cs
```diff
@@ -1,34 +1,40 @@
 ﻿using System;
```
AymericNoel, line 1: A special test project for common ! It's a good thing, we'll have other things to move additionally to NavCalculatorTests. And why not a Trakx.Persistence.test too  ?

src/Trakx.MarketData.Server/Startup.cs
```diff
@@ -68,6 +69,8 @@ public void ConfigureServices(IServiceCollection services)
             services.AddScoped<IIndiceDataProvider, IndiceDataProvider>();
             services.AddScoped<NavHub>();
             services.AddPricing();
+            services.AddSingleton<ICryptoCompareClient>(
+                new CryptoCompareClient(Environment.GetEnvironmentVariable("CRYPTOCOMPARE_API_KEY")));
```
AymericNoel, line 73: Does it work locally in we store our cryptoCompareApiKey in our .env ?

src/Trakx.Common.Tests/Unit/Pricing/NavCalculatorTests.cs
```diff
@@ -1,34 +1,40 @@
...
     {
         private readonly IMessariClient _messariClient;
         private readonly ICoinGeckoClient _coinGeckoClient;
         private readonly NavCalculator _navCalculator;
         private readonly MockCreator _mockCreator;
+        private readonly ICryptoCompareClient _cryptoCompareClient;
+        private readonly IDistributedCache _cache;
         private static readonly Task<decimal?> FailedFetchPriceResult = Task.FromResult<decimal?>(default);
 
         public NavCalculatorTests(ITestOutputHelper output)
```
jmc171144, line 30: I saw that you use this ITestOutputHelper in every test constructor now, what it is for ?

AymericNoel: it's because now MockCreator take in parameter an ITestOutputHelper

jmc171144: Haha yes but why MockCreator take it in parameter ?

src/Trakx.Common.Tests/Unit/Pricing/NavCalculatorTests.cs
monsieurleberre: Yes, let's do this little by litte. I just made 1 move to keep the PR small, but the direction is there. Persistence deserves one too, I agree.

src/Trakx.MarketData.Server/Startup.cs
monsieurleberre: Oh yes sorry, I add a variable in my launchsettings.json, this is a file which gets ignored.
image
I think .env is mostly to run integration tests right?

src/Trakx.Common.Tests/Unit/Pricing/NavCalculatorTests.cs
monsieurleberre: yes exactly it was an attempt to have random variables that stay the same on each test run, so that it is easy to debug. if you seed the generator with something deterministic, it should keep the outcome deterministic too. but we still want each test to run in its own little bubble :)

src/Trakx.MarketData.Server/Startup.cs
AymericNoel: Interesting ! I have never used these environment variable !
We should see this lauchSettingJson because I miss it ahah

PR #130, opened 2020-06-08
Dynamic Indice List page

jmc171144:
Using ApiClient for IndiceList and small refactoring


src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
```diff
@@ -1,113 +1,53 @@
...
-                new ComponentModel("MKR", 49.18),
-                new ComponentModel("wXTZ", 34.55)
-            })
-    };
-
-    public List<IndiceModel> GridData { get; set; }
-
-    #region Overrides of ComponentBase
+    public ICollection<IndiceDetailModel> GridData { get; set; }
+    private List<Object> Toolbaritems = new List<Object>() { new ItemModel() { Text = "Create Indice", TooltipText = "You will be redirected to the create indice page.", PrefixIcon = "e-add", Id = "create-indice" },
```
monsieurleberre, line 33: They are going to love it 😎

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
```diff
@@ -1,113 +1,53 @@
...
+        "Update", "ExcelExport", "PdfExport" };
 
     /// <inheritdoc />
-    protected override void OnInitialized()
+    protected override async Task OnInitializedAsync()
     {
         base.OnInitialized();
-        GridData = _indiceList;
+        GridData = (await ApiClient.GetAllIndicesAsync()).Result;
```
monsieurleberre, line 40: Cool, please make a quick test for that 🙏
Is there going to be some paging later on maybe? not for now it is fine like that 😉

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
```diff
@@ -1,113 +1,53 @@
...
-        public int ComponentCount => ComponentList.Count;
-    }
-
-    public class ComponentModel
-    {
-        public ComponentModel(string symbol, double weight)
+        if (args.Item.Id == "create-indice")
         {
-            Symbol = symbol; Weight = weight;
+            NavManager.NavigateTo("./IndiceComposition/indice_creation");
```
monsieurleberre, line 48: :ok_hand:

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
```diff
@@ -29,15 +68,24 @@
...
 
     /// <inheritdoc />
     protected override async Task OnInitializedAsync()
     {
         base.OnInitialized();
-        GridData = (await ApiClient.GetAllIndicesAsync()).Result;
+        try
+        {
+            _indiceList = (await ApiClient.GetAllIndicesAsync()).Result;
+            foreach (var indice in _indiceList) indice.IndiceCompositions = await GetComposition(indice);
```
monsieurleberre, line 82: Can you see if we can only query it when people look for the details?

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
jmc171144: The AllowPaging="true" make it automatic I think, I saw that's is possible to compute ourself the size for each page so when we'll have more index if the automatic paging doesn't suit us I would implement a function to do it correctly 😊

monsieurleberre: OK cool, I think that we also need to make sure the query behind `GetAllIndicesAsync` doesn't pull all the data at once, otherwise paging might be pointless. We'll see that later.

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
jmc171144: I test different solution this morning but anyone succeeded. I try a refresh on detail template click but the problem I'm facing is that on the first click there is no display of the compositions (I think a click is to quick comparing to data fetching). I also tried <DataManager> and the grid event DetailDataBound unsuccessfully.
I think it's possible I must be doing it wrong, but I haven't figured out how to do it yet

monsieurleberre: yes it might not be super easy, or you might need to do a call to StateHasChanged upon receiving the data to make sure the component refreshes.
It looks like the proper way to do that is to use data adaptors for WebApi like they show in here:
https://blazor.syncfusion.com/documentation/data/adaptors/

jmc171144: With StateHasChanged it works easily ✔

monsieurleberre: Ooooh yeah

PR #132, opened 2020-06-08
Adding an endpoint for currency

AymericNoel:
I added a last endpoint: CurrencyEndpoint because this endpoint allows to get the decimals of each token supported by CoinbaseCustody and thus allow the conversion of the balance and the amount of transaction in decimal and not in native token units.


src/Trakx.Coinbase.Custody.Client/Endpoints/CurrencyEndpoint.cs
```diff
@@ -0,0 +1,27 @@
...
+        {
+            _client = client;
+        }
+
+        /// <inheritdoc />
+        public async Task<PagedResponse<Currency>> ListCurrenciesAsync(string? currency = null, CancellationToken cancellationToken = default)
+        {
+            return await _client
+                .Request("currencies")
+                .SetQueryParam("symbol",currency)
```
monsieurleberre, line 23: I think it is `.AppendPathSegment()` instead
I tried and query params don't seem supported.

src/Trakx.Coinbase.Custody.Client.Tests/Unit/Endpoints/CurrencyEndpointTest.cs
```diff
@@ -0,0 +1,54 @@
...
+                .WithVerb(HttpMethod.Get);
+        }
+
+        [Fact]
+        public async Task ListCurrenciesAsync_should_deserialize_response_to_model()
+        {
+            var sampleResponse = await SampleResponseHelper.GetSampleResponseContent("PagedResponseCurrency");
+            HttpTest.RespondWith(sampleResponse);
+
+            var response = await _currencyEndpoint.ListCurrenciesAsync("btc");
```
monsieurleberre, line 42: So maybe not the best example as you pass in a symbol and return all.
Not knowing what it should do I checked manually. It looks like it should only return one currency.

src/Trakx.Coinbase.Custody.Client/Endpoints/CurrencyEndpoint.cs
AymericNoel: image

When I debug, i have the correct query parameters... are you sure ?

AymericNoel: and the tests about query parameters have succeeded

monsieurleberre: image
VS
image

I see that your code is well tested, but it is testing its own (and that is perfectly OK for a unit test). However I think it doesn't seem to match with what the outside world expects.

PR #133, opened 2020-06-09
Adding new Dao object : mappingPublicKeysDao

AymericNoel:
+  I have added new Dao object
+  New migration


AymericNoel:
> Unfortunately it is a bit too simple I think, I don't think you can link the Native addresses to the chain where they belong on the current design. I will try to do a commit to suggest better changes.

I will modify that !

src/Trakx.Persistence/Migrations/IndiceRepositoryContextModelSnapshot.cs
```diff
@@ -244,41 +244,26 @@ protected override void BuildModel(ModelBuilder modelBuilder)
...
 
-            modelBuilder.Entity("Trakx.Persistence.DAO.NativePublicKeyDao", b =>
+            modelBuilder.Entity("Trakx.Persistence.DAO.UserAddressDao", b =>
                 {
-                    b.Property<string>("PublicKey")
-                        .HasColumnType("nvarchar(450)");
-
-                    b.Property<string>("PublicKeyMappingDaoEthereumPublicKey")
+                    b.Property<string>("Id")
                         .HasColumnType("nvarchar(450)");
```
monsieurleberre, line 250: looks weird to have primary key max `450` and then user ID max `nvarchar(max)`
maybe we have that weirdness in previously generated migrations too. I never noticed...

AymericNoel: I have the impression that when we don't specify the maximum length, he automatically gives us 450.

PR #135, opened 2020-06-11
Coinbase tx stream +Tests

AymericNoel:

PR #136, opened 2020-06-12
Get last transaction date time

AymericNoel:
+  Adding minimal Datetime retrieve from last transaction in database to call CoinbaseCustody API

+  This branch is rebase from #135  so they share the same commits for base.


src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
```diff
@@ -45,7 +47,7 @@ public class CoinbaseTransactionListener : ICoinbaseTransactionListener
                         logger.LogDebug("Querying Coinbase Custody for new transactions.");
                         var page = await coinbaseClient
                             .ListTransactionsAsync(startTime:
-                                DateTime.Now.Subtract(PollingInterval.Multiply(1.1)).ToIso8601())
```
monsieurleberre: yes better!

src/Trakx.Persistence/TransactionDataProvider.cs
```diff
@@ -46,5 +48,12 @@ public async Task<List<IIndiceSupplyTransaction>> GetAllIndiceSupplyTransactions
 
             return transactions;
         }
+
+        public async Task<DateTime> GetLastWrappingTransactionDatetime()
+        {
+            var transaction = await _dbContext.WrappingTransactions.Where(t=>t.ReceiverAddress.IsValidEthereumAddress()).OrderByDescending(t=>t.TimeStamp).ToListAsync();
```
monsieurleberre, line 54: not sure you need to filter for eth address here, or maybe don't use a slow regex, but another column if you can

src/Trakx.IndiceManager.Server.Tests/Unit/Data/CoinbaseTransactionListenerTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            var observer = SimulateObservationsDuringRunningTime();
+            
+            observer.ReceivedWithAnyArgs(3).OnNext(default);
+        }
+
+        private IObserver<Transaction> SimulateObservationsDuringRunningTime()
+        {
+            var observer = Substitute.For<IObserver<Transaction>>();
+            _coinbaseListener.TransactionStream.Subscribe(observer);
+            _testScheduler.Schedule(_runningTime, s => _cancellationTokenSource.Cancel());
```
monsieurleberre, line 83: here you call _coinbaseListener.StopListening() with no token ;)

src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// <summary>
+        /// Stream of incoming transaction on Trakx coinbase wallets
+        /// </summary>
+        IObservable<Transaction> TransactionStream { get; }
+
+        /// <summary>
+        /// Allows to stop the Observable stream by passing a cancellationToken in parameter.
+        /// </summary>
+        /// <param name="cancellationToken">The CancellationToken which takes in parameter the time at which we have to stop the execution.</param>
+        void StopListening(CancellationTokenSource cancellationToken);
```
monsieurleberre, line 24: this one should not have the token as parameter, what if 2 different classes call it? then one cannot stop it anymore ;)

src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /// <inheritdoc />
+    public class CoinbaseTransactionListener : ICoinbaseTransactionListener
```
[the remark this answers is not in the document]
monsieurleberre: I agree

src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
```diff
@@ -0,0 +1,80 @@
...
+        private  CancellationTokenSource _cancellationTokenSource = new CancellationTokenSource();
+        private readonly ITransactionDataProvider _transactionDataProvider;
+
+        public CoinbaseTransactionListener(
+            ICoinbaseClient coinbaseClient, 
+            ILogger<CoinbaseTransactionListener> logger, ITransactionDataProvider transactionDataProvider, IScheduler scheduler = default)
+        {
+            _transactionDataProvider = transactionDataProvider;
+            scheduler ??= Scheduler.Default;
+            _cancellationTokenSource ??= new CancellationTokenSource();
```
monsieurleberre, line 40: replace ??= by = as this will not be passed in by constructor anymore

src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
```diff
@@ -0,0 +1,80 @@
...
+                .Distinct(t => string.Join("", t.Hashes))
+                .Do(t => logger.LogDebug("New transaction found on wallet {0}", t.WalletId));
+        }
+
+        #region Implementation of ICoinbaseTransactionListener
+
+        /// <inheritdoc />
+        public void StopListening(CancellationTokenSource cancellationToken)
+        {
+            _cancellationTokenSource = cancellationToken;
```
monsieurleberre, line 72: remove param and simply call _cancellationToken.Cancel() here

src/Trakx.Persistence/TransactionDataProvider.cs
AymericNoel: In fact, wrappingtransactions can be both wrapping or unwrapping transactions so this, I think, allow to filter only by wrapping transaction, and as it only Coinbase transaction timestamp

AymericNoel: I will add a wrap/unwrap column I think

monsieurleberre: OK it is probably worth it. One more thing is that observing a deposit and having a wrap transaction is not strictly happening at the same time. 
I think we can find a better way to issue the listener's Api requests with the correct time stamps...

AymericNoel: I totally agree !

PR #150, opened 2020-06-26
Adding new functions to query last update

AymericNoel:


src/Trakx.IndiceManager.Server.Tests/Unit/Data/BalanceUpdaterTest.cs
```diff
@@ -51,13 +51,13 @@ public void OnNext_should_creates_new_temporaryMapping_if_transactionSender_is_u
         [Fact]
         public void TryUpdateUserBalance_should_validateMapping_when_sendedAmount_is_verificationAmount()
         {
-            var retrievedUser = _daoCreator.GetRandomUserAddressDao(5);
+            var retrievedUser = _daoCreator.GetRandomUserAddressDao(-50);
             retrievedUser.IsVerified.Should().BeFalse();
-            var transaction = new Transaction{Amount = 5};
+            var processedTransaction = new ProcessedTransaction(1, new Transaction() { Amount = 5 });
```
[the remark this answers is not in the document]
jmc171144: Coday really doesn't like redundant parentheses 😂

src/Trakx.IndiceManager.Server/Data/BalanceUpdater.cs
```diff
@@ -33,20 +33,22 @@ public UserBalanceUpdater(IServiceScopeFactory serviceScopeFactory)
 
         #region Implementation of IUserBalanceUpdater
         /// <inheritdoc />
-        public bool TryUpdateUserBalance(IUserAddress retrievedUser, Transaction transaction)
+        public bool TryUpdateUserBalance(IUserAddress retrievedUser, ProcessedTransaction transaction)
         {
             var retrievedUserDao = new UserAddressDao(retrievedUser);
-            if (transaction.Amount == retrievedUserDao.VerificationAmount)
+            if (transaction.DecimalAmount == retrievedUserDao.VerificationAmount && retrievedUserDao.IsVerified==false)
```
monsieurleberre, line 39:
```suggestion
            if (transaction.DecimalAmount == retrievedUserDao.VerificationAmount && !retrievedUserDao.IsVerified)
```

src/Trakx.Persistence/UserAddressProvider.cs
```diff
@@ -25,16 +27,23 @@ public async Task<bool> ValidateMappingAddress(IUserAddress userAddressToValidat
...
             _dbContext.Entry(retrievedUser).Entity.IsVerified = true;
+            _dbContext.Entry(retrievedUser).Entity.LastUpdate = userAddressToValidate.LastUpdate;
             _dbContext.Entry(retrievedUser).Entity.Balance += retrievedUser.VerificationAmount;
             return await _dbContext.SaveChangesAsync() > 0;
         }
 
         public async Task<bool> UpdateUserBalance(IUserAddress userAddress)
         {
-            _dbContext.Entry(await _dbContext.UserAddresses.FirstOrDefaultAsync(i => i.Address == userAddress.Address)).CurrentValues.SetValues(userAddress); //modify entity of indiceDefinitionDao
+            _dbContext.Entry(await _dbContext.UserAddresses.FirstOrDefaultAsync(i => i.Address == userAddress.Address)).CurrentValues.SetValues(userAddress);
```
monsieurleberre, line 37: maybe it is safer to match on on `Id` than `Address`

src/Trakx.Persistence/UserAddressProvider.cs
```diff
@@ -25,16 +27,23 @@ public async Task<bool> ValidateMappingAddress(IUserAddress userAddressToValidat
...
             return await _dbContext.SaveChangesAsync() > 0;
         }
-        
+
+        public async Task<DateTime> GetLastTransactionDate()
+        {
+            var transactions = await _dbContext.UserAddresses.OrderByDescending(t => t.LastUpdate).ToListAsync();
```
monsieurleberre, line 43: see if you can use Max() instead, it might be optimised on the DB side

src/Trakx.Tests/Unit/Persistence/UserAddressProviderTest.cs
```diff
@@ -112,9 +112,20 @@ public async Task AddNewMapping_should_modify_User_Address_if_it_is_not_verified
             retrievedUserAddress.UserId.Should().Be("NewName");
         }
 
-        private async Task<UserAddressDao> SaveUserAddress(decimal verificationAmount=0)
+        [Fact]
+        public async Task GetLastTransactionDate_should_have_the_bigger_timestamp()
+        {
+            await SaveUserAddress(lastUpdate:DateTime.MaxValue);
```
monsieurleberre, line 118: maybe it is better to test with 2 realistic values here, I don't see the point in using a value as special as DateTime.MaxValue...

AymericNoel: At first that's what I wanted to do but then I figured the database tests could step on it and we could have conflicts if we didn't use DateTime.MaxValue.

src/Trakx.Persistence/UserAddressProvider.cs
AymericNoel: That was the plan, but I realized that if a user sends coins and he is not paired in the database, we put a temporary ID with a temporary pairing and still save the coins on his balance (This is the balanceUpdater file). Then when he wants to pair an address, normally the code makes him find his temporary pairing and his balance is not null.

monsieurleberre: OK I see that now, I think it is the fact that a `UserAddress` may not have a user that confuses me... then the Id of the class makes little sense to me too honestly...
It would be nice not to reuse the same class, but instead have another one like, ~Unknown~DepositorAddress, and the Id should really be based on the source address and the currency.

AymericNoel: Yeah an Id based on the source address and the currency may be a good thing !

PR #152, opened 2020-06-29
Wrapping mapping ui

jmc171144:
Changing the UI for wrapping according the mapping way


src/Trakx.IndiceManager.Client/Pages/Wrapping/WrapUnwrapToken.razor
```diff
@@ -1,240 +1,133 @@
...
-    /// </summary>
-    /// <returns></returns>
-    public async Task Success()
+    public SfButton SubmitButton { get; set; }
+
+    private readonly List<string> _allowedCurrencySymbols = new List<string>();
+    public static List<ComponentDetailModel>? KnownComponents { get; set; }
+    private const string EthereumAddressPattern = "^0x[a-fA-F0-9]{40}$";
+
+    [Required, RegularExpression(EthereumAddressPattern)]
```
AymericNoel, line 73: IN Trakx.Common there is an extension for dataAnnotation [EthereumAddress] something like that which that allows an easier use.

src/Trakx.IndiceManager.Client/Pages/Wrapping/WrapUnwrapToken.razor
```diff
@@ -1,240 +1,133 @@
...
-            }
-            else
-            {
-                newModel.SenderAddress = EthereumAddress;
-                newModel.ReceiverAddress = NativeAddress;
-            }
-            newModel.Date = CreationDate;
-            newModel.FromCurrency = FromCurrency;
-            newModel.ToCurrency = ToCurrency;
+    //todo: create a route that return the trakx addresses according to the chosen currency
```
AymericNoel, line 124: yes and the amount that the user must send.

src/Trakx.IndiceManager.Client/Pages/Wrapping/WrapUnwrapToken.razor
```diff
@@ -1,240 +1,133 @@
...
 
-            return newModel;
-        }
+    private void GetAmountToSend()
+    {
+        AmountToSend = (decimal)(new Random().NextDouble() / 10);
```
AymericNoel, line 128: Next, it will be send by the server.

src/Trakx.IndiceManager.Client/Pages/Wrapping/WrapUnwrapToken.razor
```diff
@@ -1,240 +1,133 @@
...
-
-        public WrappingTransactionViewModel()
+        EditContext = new EditContext(this);
+        GetAmountToSend();
+        try
         {
-            CreationDate = DateTime.Now;
-            Wrapping = true;
-            AppReceivingAddress = "Coinbase-custody-address";
+            KnownComponents = (await ApiClient.GetAllComponentsAsync()).Result;
```
AymericNoel, line 112: In fact, these are not the components you need to have here because these are not the components that users will send to coinbase. But soon there will be a route that will send you all the crypto available on coinbase.

src/Trakx.IndiceManager.Client/Pages/Wrapping/WrapUnwrapToken.razor
```diff
@@ -1,240 +1,133 @@
...
-                        </div>
+                    <div class="col-xs-12 col-sm-12 col-lg-6 col-md-6">
+                        Address to send the token to:
+                        <SfTextBox Readonly="true" @bind-Value="@ReceivingTrakxAddress" @ref="ReceivingTrakxAddressTextBox" />
                     </div>
-                    <div>
-                        <SfButton IsPrimary="true" Disabled="IsSubmitDisabled" Type="submit">@SelectedWrappingOperation</SfButton>
-                        <SfButton CssClass="e-outline" OnClick="Reset" @ref="ResetButton">Reset</SfButton>
+                    <div class="col-xs-12 col-sm-12 col-lg-6 col-md-6">
+                        Amount to send (in ETH):
```
AymericNoel, line 50: I believe the amount will be in ether to verify an ether adddress but will be in native coin for the other addresses.

src/Trakx.IndiceManager.Client/Pages/Wrapping/WrapUnwrapToken.razor
AymericNoel: Yes but Here it is the verification amount that I'm talking about, so we should send to the user the verification amount that he should sends to Trakx no ?

src/Trakx.Common/Core/UserAddress.cs
```diff
@@ -19,6 +19,16 @@ public UserAddress(string chainId, string userId, string address, decimal verifi
             CreationDate = creationDate;
         }
 
+        public UserAddress(string address, decimal verificationAmount, DateTime creationDate, bool isVerified = false, decimal balance = 0)
```
AymericNoel, line 22: this is the same constructor as above, why didn't you just pass chainId in optional ?

PR #165, opened 2020-07-08
Nav calculation for indice creation

jmc171144:


src/Trakx.IndiceManager.Client.Tests/IndiceComposition/IndiceCreationTest.cs
```diff
@@ -140,6 +144,37 @@ public void CalculateNav_should_require_one_constituent_in_model()
 
         }
 
+        [Fact]
+        public async Task CalculateNav_should_compute_correctly()
```
monsieurleberre, line 148: Cool, maybe it would be nice to have a test with more than 1 component

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -178,11 +178,22 @@
...
 
-    private decimal? CalculatedNav { get; set; }
+    public decimal? CalculatedNav { get; set; } = 0m;
+    private bool ShowCalculateNav { get; set; } = false;
+
 
     private const string _localeString = "en-US";
     private readonly List<string> _allowedConstituentSymbols = new List<string>();
-    public static List<ComponentDetailModel>? KnownComponents { get; set;}
+    public static List<ComponentDetailModel>? KnownComponents { get; set; }
```
monsieurleberre, line 187: I don't know why that one is static, that doesn't make sense to me. I knoe it is not from this PR, but can you please change it? 🙏

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -402,6 +397,108 @@
...
+        return IndiceCompositionCalculator.CalculateIndiceComposition(indiceDefinition, componentPrices,
+            TargetedNav ?? 0m, default, indiceDefinition.CreationDate);
+    }
+
+    public async Task CalculateUsdcValue()
+    {
+        try
+        {
+            var composition = await CalculateIndiceComposition();
+            var usdcValues = new Dictionary<string, decimal>();
```
monsieurleberre, line 472: now that you have the composition, could you re-use `NavCalculator.GetIndiceValuation` in order not to repeat the logic in here?
Also you will then get the ComponentValuations which can be added to the grid and presented to the user, etc...

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -402,6 +397,108 @@
...
+            await ToastError("Sorry, an error happened while trying to calculate the usdc value.");
+        }
+    }
+
+    public async Task CalculateNav()
+    {
+        try
+        {
+            await CalculateUsdcValue();
+            ConstituentRows.ForEach(o => CalculatedNav += o.Value);
```
monsieurleberre, line 493: I don't think I understand that logic, but when you press calculate nav many times, it increases the Nav each time...
image

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
jmc171144: Because the function GetComponentDetails(string Symbol) that returns the component details for a symbol need to be a static function to be call inside the ConstituentRow class

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -464,33 +519,28 @@
...
     }
 
     public async Task CalculateNav()
     {
         try
         {
-            await CalculateUsdcValue();
-            ConstituentRows.ForEach(o => CalculatedNav += o.Value);
+            var indiceValuation = await CalculateIndiceValuation();
+            CalculatedNav = 0m;
```
monsieurleberre, line 541: I think `calculatedNav = indiceValuation.NetAssetValue` or similar is possible

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -292,24 +329,38 @@
 
     public async Task SaveIndice()
     {
-        try
+        if (!IndexIsTemporary())
```
monsieurleberre, line 332: no big deal, but you can invert if to decrease nesting

src/Trakx.IndiceManager.Client.Tests/IndiceComposition/IndiceCreationTest.cs
```diff
@@ -147,32 +156,21 @@ public void CalculateNav_should_require_one_constituent_in_model()
         [Fact]
         public async Task CalculateNav_should_compute_correctly()
         {
-            SetIndexDetail();
+            var componentQuantities = new List<IComponentQuantity>{ _mockCreator.GetComponentQuantity(), _mockCreator.GetComponentQuantity() };
```
monsieurleberre, line 159: I don't know if you realise but you are setting all these objects to actually test nothing about them I think.
In essence I agree that you don't need to check the calculation itself (this is the job of `CalculateIndiceComposition`), but you need to check that you call it with the correct parameters. Same goes for `_navCalculator` call.
 -- by the way you can get a composition with many constituents using `public IIndiceComposition GetIndiceComposition(int componentCount, IIndiceDefinition? definition = default, DateTime? creationDate = default)`

monsieurleberre:
replaced by #194

PR #175, opened 2020-07-17
Reserve balances page of Coinbase custody wallets

jmc171144:


src/Trakx.Common/Extensions/DateTimeExtensions.cs
```diff
@@ -10,5 +10,12 @@ public static string ToIso8601(this DateTime dateTime)
             return dateTime.ToUniversalTime()
                 .ToString("o", CultureInfo.InvariantCulture);
         }
+
+        public static DateTimeOffset ToDateTimeOffset(this DateTime dateTime)
```
monsieurleberre, line 14: I will revert this in next commit, now that we have a better solution ;)

src/Trakx.IndiceManager.Client/Shared/NavMenu.razor
```diff
@@ -1,48 +1,80 @@
...
+        new MenuItem { Text = "My Wallet", IconCss = "fas fa-piggy-bank", Url = "user_wallet" },
+
+    };
+
+    
+    public bool IsLoggedIn = false;
+
+}
+
+<style>
```
monsieurleberre, line 50: Same I will re-delete that one ;)

src/Trakx.IndiceManager.Client/Startup.cs
```diff
@@ -23,6 +27,10 @@ public Startup(IConfiguration configuration)
         public void ConfigureServices(IServiceCollection services)
         {
             services.AddSingleton<IWeightCalculator, WeightCalculator>();
+            services.AddCoinGeckoClient();
```
monsieurleberre, line 30: I think that smells like stuff needs to move to the Server, otherwise we are doing complex stuff in what should simply be a UI 🤔

src/Trakx.MarketData.Server/Controllers/NavController.cs
```diff
@@ -64,7 +65,7 @@ public class NavController : ControllerBase
             if (!indiceOrCompositionSymbol.IsIndiceSymbol() && !indiceOrCompositionSymbol.IsCompositionSymbol())
                 return new JsonResult($"{indiceOrCompositionSymbol} is not a valid symbol.");
 
-            var utcNow = _dateTimeProvider.UtcNow;
+            var utcNow = _dateTimeProvider.UtcNow.ToDateTimeOffset();
```
monsieurleberre, line 68: I will revert this one :)

src/Trakx.Persistence.Tests/Trakx.Persistence.Tests.csproj
```diff
@@ -26,6 +26,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Trakx.Coinbase.Custody.Client\Trakx.Coinbase.Custody.Client.csproj" />
```
monsieurleberre, line 29: I am feeling weird about this one, will think about it, I don't think it is good to have a DB project depending on an external API

src/Trakx.Tests/Unit/Server/Controllers/NavControllerTests.cs
```diff
@@ -89,8 +90,8 @@ public async Task GetUsdNetAssetValue_should_return_immediately_on_bad_symbols()
         public async Task GetUsdNetAssetValue_should_retrieve_composition_asOf_utcNow_if_compositionAsOf_unspecified()
         {
             var cancellationToken = new CancellationToken();
-            var utcNow = new DateTime(2020, 02, 29);
-            var priceAsOf = new DateTime(2020, 02, 10);
+            var utcNow = new DateTimeOffset(new DateTime(2020, 02, 29)).UtcDateTime ;
```
monsieurleberre, line 93: Yes I think this is the correct fix

src/Trakx.IndiceManager.Server/Startup.cs
```diff
@@ -112,6 +118,7 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             if (env.IsDevelopment())
             {
                 app.UseDeveloperExceptionPage();
+                SeedDatabase(app);
```
monsieurleberre, line 121: this change

monsieurleberre, line 121: this change

src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
```diff
@@ -81,7 +81,7 @@ public class CoinbaseTransactionListener : ICoinbaseTransactionListener, IDispos
                 //it looks like here the only way to do a distinct would be to keep in memory all the
                 //transactions that ever appeared in the stream to perform comparison. This is probably
                 //slow and leaky.
-                .Distinct(t => string.Join("", t.Hashes))
+                .Distinct(t => string.Join("|", t.Hashes))
```
monsieurleberre, line 84: just to make sure that hashes {"abc", "de"} are different from {"ab", "cde"}

src/Trakx.IndiceManager.Server/Data/CurrencyCache.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
+using Trakx.Coinbase.Custody.Client.Interfaces;
+using Trakx.Coinbase.Custody.Client.Models;
+
+namespace Trakx.IndiceManager.Server.Data
+{
+    /// <inheritdoc />
+    public class CurrencyCache : ICurrencyCache
```
monsieurleberre, line 12: So as explained, i needed this before I realised Coinbase Custody had changed, but I don't anymore...
I think it is better to separate it anyway so I left the change in.

src/Trakx.Persistence.Tests/Trakx.Persistence.Tests.csproj
monsieurleberre: This is TEST !! so it is dirty but we need to solve this by having a separate project with only little utils at some point soon

src/Trakx.IndiceManager.Server.Tests/Unit/Data/CurrencyCacheTests.cs
```diff
@@ -0,0 +1,27 @@
+﻿using FluentAssertions;
+using NSubstitute;
+using Xunit;
+
+namespace Trakx.IndiceManager.Server.Tests.Unit.Data
```
monsieurleberre, line 5: I didnt implement it in the end as this is going to disappear when we use the new transaction model from coinbase...

src/Trakx.Coinbase.Custody.Client/Converter/StringToUlongConverters.cs
```diff
@@ -0,0 +1,50 @@
+﻿using System;
```
AymericNoel, line 1: Why do you implement JsonConverter string to Ulong or Decimal ? This is not a native function of these packages ?

src/Trakx.IndiceManager.Client/Program.cs
```diff
@@ -12,6 +13,7 @@ public static void Main(string[] args)
 
         public static IHostBuilder CreateHostBuilder(string[] args) =>
             Host.CreateDefaultBuilder(args)
+                .ConfigureAppConfiguration(b => b.AddEnvironmentVariables())
```
AymericNoel, line 16: Is this necessary to have access to environnement variables in client ? or it have an other functionnality ?

src/Trakx.IndiceManager.Client/Startup.cs
```diff
@@ -22,11 +23,14 @@ public Startup(IConfiguration configuration)
...
-            services.AddSingleton<IWeightCalculator, WeightCalculator>();
             services.AddRazorPages();
             services.AddServerSideBlazor();
             services.AddSyncfusionBlazor();
-            services.AddIndexManagerApiClient("https://localhost:5001/");
+            services.AddTransient<IToaster, Toaster>();
+            services.AddTransient<IWeightCalculator, WeightCalculator>();
+            services.AddIndexManagerApiClient(
+                Configuration.GetSection(nameof(IndexManagerApiConfiguration))
+                    .Get<IndexManagerApiConfiguration>().BaseUrl);
```
AymericNoel, line 33: I did the same in #186  but I did simplier... Why did you choose this way of implementation?

src/Trakx.IndiceManager.Server/Controllers/WrappingController.cs
```diff
@@ -80,16 +96,20 @@ public async Task<ActionResult<List<WrappingTransactionModel>>> GetAllTransactio
...
         /// <summary>
-        /// This route allows to retrieve the Trakx's balance, either with native or wrapped tokens.
+        /// This route allows to retrieve the Trakx' balance, either with native or wrapped tokens.
         /// </summary>
-        /// <returns>An Ok object list composed by all the balances with status 200 or a BadRequest Object if it failed.</returns>
+        /// <returns>An Ok object list composed by all the balances with status 200 or a InternalServerError object if it failed.</returns>
         [HttpGet]
-        public async Task<ActionResult<List<AccountBalanceModel>>> GetTrakxBalance()
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
```
AymericNoel, line 104: Yeah you add this everywere, it's cool !

src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
```diff
@@ -32,26 +31,27 @@ public interface ICoinbaseTransactionListener
 
     /// <inheritdoc cref="ICoinbaseTransactionListener"/>
     /// <inheritdoc cref="IDisposable"/>
-    public class CoinbaseTransactionListener : ICoinbaseTransactionListener, IDisposable
+    public sealed class CoinbaseTransactionListener : ICoinbaseTransactionListener, IDisposable
```
AymericNoel, line 34: Why did you pass this class to ```sealed class``` ?

src/Trakx.IndiceManager.Server/Managers/WrappingService.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        /// <inheritdoc />
+        public async IAsyncEnumerable<AccountBalanceModel> GetTrakxBalances(
+            [EnumeratorCancellation] CancellationToken cancellationToken = default)
+        {
+            IAsyncEnumerable<Wallet> wallets;
+            try
+            {
+                wallets = _coinbaseClient.GetWallets(cancellationToken: cancellationToken);
```
AymericNoel, line 62: This returns all the wallets but we'll return the users's wallets too (If we create a wallet per user). This is not a problem ?

src/Trakx.Persistence.Tests/Trakx.Persistence.Tests.csproj
AymericNoel: Agree

src/Trakx.IndiceManager.Server/Controllers/WrappingController.cs
```diff
@@ -46,10 +56,12 @@ public async Task<ActionResult<string>> GetTrakxAddressFromSymbol([FromBody]stri
         /// </summary>
         /// <param name="transaction">The <see cref="WrappingTransactionModel"/> which that has all the information needed to
         /// complete the transaction.</param>
-        /// <returns>Return a BadRequest object if the informations in <paramref name="transaction"/> are wrong.
+        /// <returns>Return a BadRequest object if the information in <paramref name="transaction"/> are wrong.
         /// Else it return an OK object result.
         /// </returns>
         [HttpPost]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
```
jmc171144, line 66: These make the rejected errors cleaner but when you are debugging you don't get details of the error, and it can be a pain in the ass when you're trying to debug 😥

src/Trakx.IndiceManager.Client/Shared/IToaster.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Trakx.IndiceManager.Client.Shared
+{
+    public interface IToaster
+    {
+        Task ShowInfo(string content, string title = "Info");
+        Task ShowSuccess(string content, string title = "Success");
+        Task ShowWarning(string content, string title = "Warning");
+        Task ShowError(string content, string title = "Error");
+    }
```
jmc171144, line 11: Thanks 🙏

src/Trakx.IndiceManager.Client/Shared/Toaster.razor
```diff
@@ -0,0 +1,63 @@
...
+        {
+            Title = title,
+            Content = content,
+            CssClass = toastClass,
+            Icon = icon
+        }));
+    }
+
+}
+
```
jmc171144, line 62: Nice! it'll be a lot cleaner that way

src/Trakx.IndiceManager.Server/Controllers/WrappingController.cs
monsieurleberre: I agree, but I removed the error messages  because it is usually considered a vector of attack to give too much info on what your are failing for...

src/Trakx.IndiceManager.Server/Managers/WrappingService.cs
monsieurleberre: I am not sure it is, in one case, we return all the wallets that Trakx owns, with a link to the user, in another we return a single wallet where transaction history can be filtered by the source of the funds. So in both cases you can investigate to see which user has how much funds. We have to think about it...

src/Trakx.IndiceManager.Server/Data/CoinbaseTransactionListener.cs
monsieurleberre: Just so that we don't have to make the implementation of IDisposable inheritance friendly. --meaning having a protected virtual IDisposable(bool isDisposing)

src/Trakx.IndiceManager.Server/Controllers/WrappingController.cs
monsieurleberre: just on the one I was changing ;) there is still a bit left everywhere

src/Trakx.IndiceManager.Client/Startup.cs
monsieurleberre: ah let me see... Here you can have it from an environment variable called `IndexManagerApiConfiguration__BaseUrl`, or in the `AppSettings.json` so it seems to me it is both docker and local dev friendly

src/Trakx.IndiceManager.Client/Program.cs
monsieurleberre: this is so you can pass things easily when running docker 🤔 is this an issue anywhere?

src/Trakx.Coinbase.Custody.Client/Converter/StringToUlongConverters.cs
monsieurleberre: I wish it was honestly, but it didn't seem to be (or I missed something??) so you end up with strings that are polluting the code downstream... Also coinbase custody is so great that they write some integers like `234.0` which is removed in the ULong converter 🙉

src/Trakx.IndiceManager.Client/Startup.cs
AymericNoel: Okay I see !

PR #186, opened 2020-07-22
Repare local docker-compose

AymericNoel:


AymericNoel:
The docker-compose is running locally but we can't see our pages because, we have still problems with Nginx I think but it's better than before, at least the file's names are corrects

src/docker-compose.local.yml
```diff
@@ -24,29 +24,49 @@ services:
...
     build:
       context: ./..
-      dockerfile: ./src/IndiceManager.Dockerfile
+      dockerfile: ./src/IndexManagerServer.Dockerfile
     links:
       - redis
       - sqlserver
     environment:
-      - ASPNETCORE_ENVIRONMENT=Development #Consider changing this in Production
+      - ASPNETCORE_ENVIRONMENT=Production #Consider changing this in Production
```
monsieurleberre, line 40: that won't give you detailed errors on local, did you really want to do that?

AymericNoel: in this docker-compose, we have marketData and IndexManager, if they're both in development, they'll try to seed Database at the same time, that's why I put indexManager in prod, to avoid duplicate seeding.

What do you tkink?

monsieurleberre: Oh yes of course I forgot about that 🤦‍♂️🤦‍♂️I guess it is fine, alternatively we could do a bit of config to say who does it but keep everybody in dev mode? As you wish :)

AymericNoel: Agree !

AymericNoel: To do a bit of config

Issue #185, opened 2020-07-21
Adapt Coinbase custody API to use new improved transaction objects


monsieurleberre:
That means we can get rid of
1) CurrencyCache
2) CoinbaseRawTransaction
and simplify CoinbaseTransactionListener

PR #189, opened 2020-07-22
Updating coinbase package with their new version

AymericNoel:
links to #185


src/Trakx.Coinbase.Custody.Client/Models/CoinbaseTransaction.cs
```diff
@@ -1,21 +1,145 @@
...
+        [JsonProperty("type")]
+        [JsonPropertyName("type")]
+        public TransactionType Type { get; set; }
+
+        [JsonProperty("state")]
+        [JsonPropertyName("state")]
+        public TransactionState State{ get; set; }
+
+        [JsonProperty("amount")]
+        [JsonPropertyName("amount")]
```
monsieurleberre, line 27: So it looks like I didn't really need those decimal converters on the wallet 🤔 damned what a waste of time !!

src/Trakx.IndiceManager.Server.Tests/Unit/Controllers/IndiceCreationControllerTests.cs
```diff
@@ -300,7 +300,7 @@ public async Task SaveComponentDefinition_should_send_badRequestError_if_compone
             await _componentRetriever.DidNotReceiveWithAnyArgs().TryToSaveComponentDefinition(component);
         }
 
-        [Fact]
+        [Fact(Skip = "cannot run with other tests due to conflicts.")]
```
monsieurleberre, line 303: Is that one that fails from time to time?

src/Trakx.IndiceManager.Server.Tests/Unit/Data/BalanceUpdaterTest.cs
```diff
@@ -37,13 +37,12 @@ private IServiceScopeFactory PrepareScopeResolution()
...
         {
-            var transaction = new CoinbaseTransaction(
-                    new CoinbaseRawTransaction
-                    {
-                        Currency = _daoCreator.GetRandomString(3), 
-                        Source = _daoCreator.GetRandomAddressEthereum(),
-                        UnscaledAmount = _daoCreator.GetRandomUnscaledAmount(),
-                    }, _daoCreator.GetRandomNaturalUnit());
+            var transaction = new CoinbaseTransaction
+            {
```
monsieurleberre, line 41: Maybe a new method in the coinbase custody mockcreator?

PR #191, opened 2020-07-24
rename depositaddress to external address as they should support withdrawals too

monsieurleberre:
I started to do a PR for user wallets (in the sense of Coinbase Custody) but this first step is already good enough for a PR
It is just plain and simple renaming :)


src/Trakx.IndiceManager.ApiClient/Trakx.IndiceManager.ApiClient.csproj
```diff
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
-    <GenerateApiClient>False</GenerateApiClient>
+    <GenerateApiClient>True</GenerateApiClient>
```
monsieurleberre, line 5: oops!! Let me turn it back to false

PR #194, opened 2020-07-27
Nav calculation for indice creation from server

jmc171144:


src/Trakx.IndiceManager.Client.Tests/IndiceComposition/IndiceCreationTest.cs
```diff
@@ -268,6 +348,36 @@ public async Task WeightRefresh_for_unit_in_percentage_should_calculate_usdc_val
                 , Arg.Is<decimal>( n => n == (_componentInstance.TargetedNav ?? 0m)));
         }
 
+        [Fact]
+        public async Task CalculateIndiceComposition_should_work_as_expected()
+        {
+            var componentDetail = MockComponentDetailConversion();
```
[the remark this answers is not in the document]
jmc171144: 💩y test

src/Trakx.IndiceManager.ApiClient/Trakx.IndiceManager.ApiClient.csproj
```diff
@@ -15,6 +15,7 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Trakx.Common\Trakx.Common.csproj" />
+    <ProjectReference Include="..\Trakx.IndiceManager.Server\Trakx.IndiceManager.Server.csproj" />
```
monsieurleberre, line 18: OK I see why you need that since you have a new IModel conversion, but the whole point of this client/server thing is not to couple the client and the server (apart from Api methods, which are obviously going to have some common contract to respect if client and server are to communicate...).
If we were ok to reference the server in the apiclient, and consequently in the client, we would have generated razor pages directly from the server.
Here you are asking the user of the client to get all the dlls involved in the server, just to translate little DTO objects.
As commented on the mockCreator in common, the only alternative is to have a separate project only for the DTOs, or Models as they are named in the current solution. Then you can reference this little project in both and share the models, but _only_ the models.

src/Trakx.IndiceManager.Client.Tests/IndiceComposition/IndiceCreationTest.cs
```diff
@@ -31,15 +42,23 @@ public IndiceCreationTest(ITestOutputHelper output)
 
             _apiClient = Substitute.For<IIndiceCreationClient>();
             _weightCalculator = Substitute.For<IWeightCalculator>();
+            _navCalculator = Substitute.For<INavCalculator>();
```
monsieurleberre, line 45: I am not sure this is what we discussed, you should only need to mock the stuff that lives on client side, navCalculator for instance should not be here.

src/Trakx.IndiceManager.Client.Tests/IndiceComposition/IndiceCreationTest.cs
```diff
@@ -156,6 +216,26 @@ public async Task ViewModel_should_be_empty_after_saved()
             _componentInstance.IndexDescription = indiceDefinition.Description;
             Component.Instance.ConstituentRows.Add(_constituent);
 
+            var componentQuantity = _mockCreator.GetComponentQuantity();
+            var weight = _mockCreator.GetRandomWeight();
+            var detail = new ComponentDetailModel(componentQuantity, weight);
```
monsieurleberre, line 221: Here i'd use GetRandomComponentDetailModel() from the Trakx.IndiceManager.Client.Tests.MockCreator

src/Trakx.IndiceManager.Client.Tests/IndiceComposition/IndiceCreationTest.cs
```diff
@@ -11,8 +11,15 @@
...
 using System.Linq;
+using NSubstitute.ExceptionExtensions;
 using Trakx.Common.Composition;
+using Trakx.Common.Core;
+using Trakx.Common.Interfaces.Indice;
+using Trakx.Common.Interfaces.Pricing;
+using Trakx.Common.Pricing;
+using Trakx.Common.Sources.CoinGecko;
 using Trakx.IndiceManager.ApiClient;
+using ComponentDetailModel = Trakx.IndiceManager.Server.Models.ComponentDetailModel;
```
monsieurleberre, line 22: I don't think you picked the good one, here you are testing client side, and going through a lot of pain with server side models

src/Trakx.Persistence/IndiceRepositoryContext.cs
```diff
@@ -5,8 +5,21 @@
 using Trakx.Common.Interfaces;
 
 namespace Trakx.Persistence
-{
-    public partial class IndiceRepositoryContext : DbContext
+{
+    public interface IIndiceRepositoryContext
```
monsieurleberre, line 9: I have nothing agains that (actually why adding setters?), but where was it needed?

src/Trakx.IndiceManager.Server.Tests/Trakx.IndiceManager.Server.Tests.csproj
```diff
@@ -31,6 +31,7 @@
   <ItemGroup>
     <ProjectReference Include="..\Trakx.Coinbase.Custody.Client.Tests\Trakx.Coinbase.Custody.Client.Tests.csproj" />
     <ProjectReference Include="..\Trakx.Common.Tests\Trakx.Common.Tests.csproj" />
+    <ProjectReference Include="..\Trakx.Exchange.ApiClient\Trakx.Exchange.ApiClient.csproj" />
```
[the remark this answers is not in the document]
jmc171144: It's only for using the ApiException in tests, and the one in IndiceManager.ApiClient creates confusing name errors. But I can create a new ApiException class if you prefer not involved too much projects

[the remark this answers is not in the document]
monsieurleberre: I don't think there should be a test that the client receives an ApiException in the server tests honestly. You can test that a BadRequest is sent, but not how the client should react to it...

PR #199, opened 2020-08-04
db background service for IndiceValuation calculation on a daily basis

jmc171144:


jmc171144:
I am not using this service anymore because consuming scoped service seems more appropriate for calling async methods

AymericNoel:
I see that you use ```await Task.delay``` and a while loop, why don't you use observable as we did for ```CoinbaseTransactionListener``` ?

src/Trakx.MarketData.Server/MyBackgroundService.cs
```diff
@@ -0,0 +1,108 @@
...
+using Trakx.Common.Interfaces.Pricing;
+
+namespace Trakx.MarketData.Server
+{
+    internal interface IMyBackgroundService
+    {
+        Task DoWork(CancellationToken stoppingToken);
+    }
+
+    internal class MyBackgroundService : IMyBackgroundService
```
monsieurleberre, line 17: OK so it is time to start testing what this service does I think.
Anything that you didn't declare on the interface shouldn't be tested, and should be made private.
You need to test what happens to DoWork under different conditions.

src/Trakx.MarketData.Server/Data/DataUpdater/DatabaseUpdaterExecutor.cs
```diff
@@ -25,21 +26,21 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
...
             _logger.LogInformation(
-                "Background Service Hosted Service is working.");
+                "Hourly basis indice valuations refreshment is working.");
 
             using (var scope = Services.CreateScope())
             {
                 var scopedProcessingService =
                     scope.ServiceProvider
-                        .GetRequiredService<IMyBackgroundService>();
+                        .GetRequiredService<IIndiceValuationsUpdater>();
```
monsieurleberre, line 42: I would instantiate the observer here, and do the subscription to the Valuations stream just after, not in the RefreshValuationsEveryHours.
Otherwise you cannot mock the `new IndiceValuationsObserver()` that you are using in `RefreshValuationsEveryHours` and check how it is being called.

src/Trakx.MarketData.Server/Data/DataUpdater/DatabaseUpdaterExecutor.cs
```diff
@@ -25,21 +26,21 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             _logger.LogInformation(
                 "Background Service Hosted Service running.");
 
-            await DoWork(stoppingToken);
+            await ExecuteValuationsRefreshment(stoppingToken).ConfigureAwait(false);
         }
 
-        private async Task DoWork(CancellationToken stoppingToken)
+        private async Task ExecuteValuationsRefreshment(CancellationToken stoppingToken)
```
monsieurleberre, line 32: or... RefreshValuations, is a bit simpler ;) but then I would rather call it `PersistValuations`

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationsUpdater.cs
```diff
@@ -0,0 +1,176 @@
...
+
+namespace Trakx.MarketData.Server.Data.DataUpdater
+{
+    public interface IIndiceValuationsUpdater
+    {
+        /// <summary>
+        /// Computes all the valuations for the indice retrieved from the database.
+        /// </summary>
+        /// <returns>List of indice valuations.</returns>
+        IAsyncEnumerable<IIndiceValuation> ComputeAllValuations();
```
monsieurleberre, line 24: I don't think this should be on the interface at all, it is merely internal logic...
I am not sure if you did it for testing but as per this comment (https://github.com/trakx/trakx-tools/pull/199#discussion_r465533674) this would be bad practice.

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationsUpdater.cs
```diff
@@ -71,13 +86,10 @@ public async Task RefreshValuationsEveryHours(CancellationToken stoppingToken)
...
                     t.IndiceComposition.Symbol, t.NetAssetValue, t.TimeStamp);
                 });
-
-            Valuations.Subscribe(new IndiceValuationsObserver());
         }
 
 
-
-        private async IAsyncEnumerable<IIndiceValuation> ComputeAllValuations()
+        public async IAsyncEnumerable<IIndiceValuation> ComputeAllValuations()
```
monsieurleberre, line 92: this is internal logic and shouldn't encumber the interface.

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationsUpdater.cs
```diff
@@ -106,7 +111,6 @@ private async Task<List<string>> GetAllSymbols()
             }
             catch (Exception e)
             {
```
monsieurleberre, line 113: it is good to log errors, although they should be logged with the correct level, and the exception should be passed with the correct method signature ie `logger.LogError(e, "some message giving context")
When you are returning values to an external consumer, you shouldn't include details about where you are failing, but here this is usefull for our own diagnosis.

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationsUpdater.cs
```diff
@@ -142,20 +145,29 @@ private async Task<List<IIndiceValuation>> GetAllValuations(List<IIndiceComposit
...
+            _cancellationTokenSource.Cancel();
+        }
+
+
         #region Implementation of IDisposable
 
         /// <inheritdoc />
         public void Dispose()
         {
             _cancellationTokenSource?.Dispose();
```
monsieurleberre, line 170: I actually think that you should dispose of the scope when you are finished using this class so that the connection to the database can be closed.

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationsUpdater.cs
jmc171144: Yes I deleted it when I moved all these functions in another service but I forgot to put it back

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationsUpdater.cs
jmc171144: It cannot be here actually because it is generating an error, or something have to change elsewhere

monsieurleberre: I would say something has to change elsewhere then!

src/Trakx.MarketData.Server/Data/DataUpdater/DatabaseUpdaterExecutor.cs
```diff
@@ -0,0 +1,55 @@
...
+            _logger.LogInformation(
+                "Hourly basis indice valuations refreshment is working.");
+
+            using (var scope = Services.CreateScope())
+            {
+                var scopedProcessingService =
+                    scope.ServiceProvider
+                        .GetRequiredService<IIndiceValuationsUpdater>();
+                
+                scopedProcessingService.ValuationStream.Subscribe(new IndiceValuationsObserver());
```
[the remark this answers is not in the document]
jmc171144: For 2) yes for the moment the class IndiceValuationsObserver is empty, but if later I add in it the functionalities to persist -when it will be working- it will be ok, won't it ?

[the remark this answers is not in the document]
monsieurleberre: yes yes it is just slightly adding to the confusion of the whole thing: You seem to have built a service meant to publish new IndexValuation elements onto a stream, then you have a service that you call `DatabaseUpdaterExecutor` which should simply react to new IndexValuations and execute the corresponding DB Update (if it is named correctly, and/or if I follow your intentions). So this service is the observer of what is coming onto the stream. Another service could be doing something else, like checking the level of the Nav on the stream and sending a trading alert.

So, given that many components might be interested in that stream, what should link your `Streamer`, ie. the thing that _produces_ IndexValuations updates, to the services interested by these updates, is a long running background service, which is in only charge of subscribing the observers onto the stream, and maintaining those subscription throughout all its lifecycle.

Currently, you don't seem to have separated clearly in the code, and probably in your head, the role of each class, and the life expectancy of its dependencies. I am going to be boring 🥱 and say that starting with good names, or even not starting if you can't give at least decent names, would help you to make this a bit cleaner.

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdater.cs
```diff
@@ -74,27 +73,27 @@ public class IndiceValuationsUpdater : IIndiceValuationsUpdater, IDisposable
...
-                var compositions = await GetAllCompositions(symbols);
-                valuations = await GetAllValuations(compositions);                
+                var symbols = await GetAllSymbols().ConfigureAwait(false);
+                var compositions = await GetAllCompositions(symbols).ConfigureAwait(false);
+                valuations = await GetAllValuations(compositions).ConfigureAwait(false);                
             }
             catch(Exception e)
             {
-                _logger.LogInformation(e.Message);
+                _logger.LogError(e, "An error happened when trying to copute valuations from compositions in database.");
```
monsieurleberre, line 96: ah yes this is the way to log!

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdater.cs
```diff
@@ -169,6 +166,7 @@ public void StopListening()
         public void Dispose()
         {
             _cancellationTokenSource?.Dispose();
+            _initialisationScope?.Dispose();
```
monsieurleberre, line 169: yes i think i agree 👍

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdaterService.cs
```diff
@@ -0,0 +1,87 @@
...
+        public IndiceValuationUpdaterService(IServiceProvider services,
+            ILogger<IndiceValuationUpdaterService> logger,
+            IIndiceValuationWriter valuationWriter)
+        {
+            Services = services;
+            _logger = logger;
+            _valuationWriter = valuationWriter;
+            _serviceScope = Services.CreateScope();
+        }
+        public IndiceValuationUpdaterService(IIndiceValuationWriter valuationWriter)
```
monsieurleberre, line 30: should we only keep the first constructor? otherwise the first should call the second to do the assignment of `valuationWriter`

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdaterService.cs
```diff
@@ -0,0 +1,87 @@
...
+using System.Linq;
+using System.Reactive;
+using System.Reactive.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Trakx.Common.Interfaces.Indice;
+
+namespace Trakx.MarketData.Server.Data.DataUpdater
+{
+    public class IndiceValuationUpdaterService : BackgroundService, IObserver<IIndiceValuation>
```
monsieurleberre, line 15: so there is still a bit of confusion I think; the IIndiceValuationWriter is the one interested in updates, so he is the observer.
The background service will only subscribe its `TryToSaveValuation` to the stream coming from IndiceValuation when startasync or executeasync gets called.

I think there should be a long running thing in ExecuteAsync that can be stopped by a cancellation token. Otherwise the long running task will terminate and your scopes will be gone.

Please have a look and try to understand what is going on in `BalanceUpdaterService`, your task is very very similar ;)

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdaterService.cs
```diff
@@ -12,34 +12,34 @@
...
         #region IHostedService implementation
         protected override async Task ExecuteAsync(CancellationToken stoppingToken)
-        {
-            _logger.LogInformation(
-                "Background Service IndiceValuationUpdaterService is running.");
-
+        {            
             await PersistValuations().ConfigureAwait(false);
+
+            while (!stoppingToken.IsCancellationRequested)
```
monsieurleberre, line 37: haha! now your stuff might not get disposed instantly anymore 😜

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdaterService.cs
```diff
@@ -61,27 +61,11 @@ private async Task PersistValuations()
                 _serviceScope.ServiceProvider
                 .GetRequiredService<IIndiceValuationUpdater>();
 
-            scopedProcessingService.ValuationStream.Subscribe(new IndiceValuationUpdaterService(_valuationWriter));
+            scopedProcessingService.ValuationStream.Subscribe(_observer);
```
monsieurleberre, line 64: yes! so here you could decide to inject a collection of IObsever<IIndiceValuation> in the constructor and subscribe each of them to the stream, maybe one does logging, one does persistence, one does some trading notifications to slack, one is a rest call to publish the valuation on a dashboard somewhere, etc, etc... Hopefully you can see that it would be quite easy with this construction to have many different ways to consume valuations, without changing much code!

jmc171144: Yes I see! It would be much simpler indeed

src/Trakx.Marketdata.Server.Tests/Unit/Data/IndiceValuationsUpdaterTests.cs
```diff
@@ -0,0 +1,133 @@
...
+        }
+
+        [Fact]
+        public void ValuationStream_should_not_stop_if_NavCalculator_return_an_error()
+        {
+            _navCalculator.GetIndiceValuation(default).ThrowsForAnyArgs(new Exception());
+
+            SimulateObservationsDuringRunningTime();
+
+            _dataProvider.ReceivedWithAnyArgs(10).GetAllIndiceSymbols();
```
monsieurleberre, line 111: You can reuse CheckAllReceivedCalls right?

src/Trakx.Marketdata.Server.Tests/Unit/Data/IndiceValuationsUpdaterTests.cs
```diff
@@ -0,0 +1,133 @@
...
+    public class IndiceValuationsUpdaterTests
+    {
+        private readonly IIndiceDataProvider _dataProvider;
+        private readonly INavCalculator _navCalculator;
+        private readonly IIndiceValuationUpdater _valuationsUpdater;
+        private readonly MockCreator _mockCreator;
+        private readonly TestScheduler _testScheduler;
+        private readonly TimeSpan _runningTime;
+
+        public IndiceValuationsUpdaterTests(ITestOutputHelper output)
```
monsieurleberre, line 34: Nice 👍

PR #204, opened 2020-08-11
run the tests to get fake trade data from the past

monsieurleberre:
part of resolution of https://github.com/trakx/exchange-config-prod/issues/26


src/Trakx.Drafts/Tools/FakeTradeHistoryGenerator.cs
```diff
@@ -0,0 +1,123 @@
...
+        public FakeTradeHistoryGenerator(SeededDbContextFixture fixture)
+        {
+            _context = fixture.Context;
+
+            var serviceCollection = new ServiceCollection();
+
+            serviceCollection.AddPricing();
+            serviceCollection.AddMemoryCache();
+            serviceCollection.AddMessariClient();
+            serviceCollection.AddCoinGeckoClient();
```
jmc171144, line 41: 🤔 why all services are added directly in the class constructors in this project ?

monsieurleberre: So this one is more like an integration test than a unit test. It is actually not even a test, but just a way to conveniently run a custom bit of code.
In that sense, it is almost like when you start the server, you tell it all the services you are going to need to do the job, and then perform actions with real classes, not mocked ones.
Since I need real prices, I have to get the real compositions, then get the real data providers => I have a functional NavCalculator, and I can finally use it to generate a CSV file.

PR #205, opened 2020-08-12
Allows to add composition on existing index or allows to rebalance

AymericNoel:


src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceList.razor
```diff
@@ -7,18 +7,19 @@
 @using Trakx.Common.Models
 @using Trakx.IndiceManager.ApiClient
 @using Trakx.IndiceManager.Models
-@inject IIndiceCreationClient ApiClient;
+@inject IIndiceCreationClient ApiClient
 @inject NavigationManager NavManager
 
-<SfToast @ref="Toaster" TimeOut=3000>
+<SfToast @ref="Toaster" TimeOut=3000 ShowCloseButton="true" ShowProgressBar="true">
```
monsieurleberre, line 13: if you are going to touch this one, can you please try to reuse `Trakx.IndiceManager.Client.Shared.Toaster` ? It should be more convenient to test and less verbose in code

src/Trakx.IndiceManager.Client/appsettings.json
```diff
@@ -8,6 +8,6 @@
   },
   "AllowedHosts": "*",
   "IndexManagerApiConfiguration": {
-    "BaseUrl": "https://localhost:44376/"
+    "BaseUrl": "https://localhost:5001"
```
monsieurleberre, line 11: yes for me too that is better 😃

src/Trakx.IndiceManager.Server/Controllers/IndiceCreationController.cs
```diff
@@ -100,7 +100,7 @@ public async Task<ActionResult<ComponentDetailModel>> SaveComponentDefinition(Co
         /// </summary>
         /// <returns>A list of indices with details about them.</returns>
         [HttpGet]
-        [Authorize(Roles = "admin")]
+        [Authorize(Roles = "Admin")]
```
monsieurleberre, line 103: Admin and SuperAdmin? or am I missing something?

src/Trakx.IndiceManager.Server/Managers/IndiceInformationRetriever.cs
```diff
@@ -41,5 +41,17 @@ public async Task<bool> SearchCompositionByAddress(string compositionAddress)
         {
             return await _dataProvider.TryToGetCompositionByAddress(compositionAddress);
         }
+
+        /// <inheritdoc />
+        public async Task<IIndiceComposition?> GetCompositionFromSymbol(string compositionSymbol)
```
monsieurleberre, line 46: I am wondering why this class exists, if it seems to be only doing IIndiceDataProvider's job, could we get rid of it, it seems useless to maintain this extra code...

src/Trakx.Common/Extensions/NavigationManagerExtension.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.AspNetCore.Components;
+using Microsoft.AspNetCore.WebUtilities;
+
+namespace Trakx.Common.Extensions
+{
+    public static class NavigationManagerExtensions
```
monsieurleberre, line 9: 0 references to that code so far 😅 I will try to understand it in context later ;)

src/Trakx.IndiceManager.Server/Managers/IndiceInformationRetriever.cs
AymericNoel: Ok looks good too !

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -312,7 +313,60 @@
...
+                IndexSymbol = retrievedComposition.IndiceDetail.Symbol;
+                retrievedComposition.Components.ForEach(async c =>
+               {
+                   var row = new ConstituentRow()
+                   {
+                       DetailModel = c,
+                       Symbol = c.Symbol,
+                       CustomWeight = c.Weight ?? 1m,
+                       Value = 10m
+                   };
```
jmc171144, line 334: Why set theses values ? And by the way all components cannot have a weight equal to 1, is there something I didn't understand?

AymericNoel:
So : 
+ New feature : We can add just an index without having a composition on the indice-creation page
+ If the index is published (cf. he has an address) it is readonly on the indice-creation page (just if we load a composition, doesn't work when we load only an index, I don't know why yet)
+ I have deleted the temporary-address when saving a composition or an index in the database because when they are loaded on the indice-list page, they have the "published" state because they have an address, and they become readonly on the indice-creation page
+ The price, quantity and weight are loaded on the indice-creation page

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/AddERC20.razor
```diff
@@ -89,7 +88,7 @@
         catch (ApiException ex)
         {
             ErrorMessage = ex.Response;
-            await InvokeAsync(() =>ToastObj.Show(new ToastModel { Title = "Error!", Content = $"{ErrorMessage}", CssClass = "e-toast-danger" }));
+            await InvokeAsync(() => SharedToaster.SfToaster.Show(new ToastModel { Title = "Error!", Content = $"{ErrorMessage}", CssClass = "e-toast-danger" }));
```
monsieurleberre, line 91: can't we use `IToaster.ShowError` ? Otherwise we sort of miss the point of not repeating code I think

src/Trakx.IndiceManager.Client/Pages/IndiceComposition/IndiceCreation.razor
```diff
@@ -306,62 +285,61 @@
...
+                        Price = c.Price
+                    };
+                    ConstituentRows.Add(row);
                 });
                 TargetedNav = retrievedComposition.TargetedNav;
-
                 RefreshWeights();
             }
-            catch (Exception)
+            catch (ApiException)
```
monsieurleberre, line 322: I think it was not too bad to catch all errors, otherwise the whole frontend seems to crash (I have seen that debugging before, but not sure how much you tested this one)

src/Trakx.IndiceManager.Client/Shared/IToaster.cs
```diff
@@ -8,5 +9,7 @@ public interface IToaster
         Task ShowSuccess(string content, string title = "Success");
         Task ShowWarning(string content, string title = "Warning");
         Task ShowError(string content, string title = "Error");
+
+        public SfToast SfToaster { get; set; }
```
monsieurleberre, line 13: I am not sure I want to make that public, the aim of the class was initially to hide a bit of complexity, but where do you need it? Also, if you really have to expose it public, don't add the `public` keyword in an interface, and expose the interface from which SfToast is derived.

src/Trakx.IndiceManager.Server.Tests/Unit/Controllers/IndiceCreationControllerTests.cs
```diff
@@ -449,18 +453,79 @@ public async Task CalculateNavFromComposition_should_return_Conflict_if_net_asse
...
+            _navCalculator.GetIndiceValuation(default).ReturnsForAnyArgs(valuation);
+
+            var response = await _controller.CalculateNavFromComposition(composition);
+            var netAssetValue = ((OkObjectResult)response.Result).Value;
+
+            netAssetValue.Should().Be(valuation.NetAssetValue);
         }
 
+        [Fact]
+        public async Task GetIndexCompositionFromSymbol_should_return_error_if_symbol_empty()
```
monsieurleberre, line 470: can you check for the error or change the test name then 😉 thanks!

src/Trakx.IndiceManager.Server.Tests/Unit/Controllers/IndiceCreationControllerTests.cs
```diff
@@ -449,18 +453,79 @@ public async Task CalculateNavFromComposition_should_return_Conflict_if_net_asse
...
+            retrievedComposition.TargetedNav.Should().Be(valuation.NetAssetValue);
+            retrievedComposition.Components[0].Quantity.Should().Be(valuation.ComponentValuations
+                .First(c => c.ComponentQuantity.ComponentDefinition.Symbol == retrievedComposition.Components[0].Symbol)
+                .ComponentQuantity.Quantity);
+            retrievedComposition.Components[1].Weight.Should().Be(Convert.ToDecimal(valuation.ComponentValuations
+                .First(c => c.ComponentQuantity.ComponentDefinition.Symbol == retrievedComposition.Components[1].Symbol).Weight));
+        }
+
+        [Fact]
+        public async Task GetIndexDefinitionFromSymbol_should_return_error_if_symbol_empty()
```
monsieurleberre, line 505: same here maybe 🤷‍♂️

PR #208, opened 2020-08-14
Persist indice valuation in db

jmc171144:


src/Trakx.Persistence/IndiceDataCreator.cs
```diff
@@ -38,17 +37,67 @@ public async Task<bool> AddNewComposition(IIndiceComposition indiceComposition)
...
+                    .Single(o => o.ComponentQuantity.ComponentDefinition.Symbol == attachedQuantity.ComponentDefinition.Symbol);
+
+                var attachedValuation = new ComponentValuationDao(attachedQuantity,
+                    componentValuation.TimeStamp, componentValuation.QuoteCurrency,
+                    componentValuation.Price, componentValuation.PriceSource);
+
+                attachedValuations.Add(attachedValuation);
+            }
+
+            var valuationDao = new IndiceValuationDao(indiceValuation)
```
monsieurleberre, line 85: Can you please try to use this constructor `public IndiceValuationDao(List<ComponentValuationDao> componentValuations)` ?

src/Trakx.Persistence/DAO/IndiceValuationDao.cs
```diff
@@ -30,6 +31,29 @@ public IndiceValuationDao(List<ComponentValuationDao> componentValuations)
             Id = $"{indiceComposition.Id}|{quoteCurrency}|{timeStamp:yyMMddHHmmssff}";
         }
 
+        public IndiceValuationDao(IIndiceValuation indiceValuation)
```
monsieurleberre, line 34: if adding this constructor, can you please try to re-use the other constructors which do a bit of work already?

src/Trakx.Persistence/IndiceDataCreator.cs
```diff
@@ -68,35 +69,34 @@ public async Task<bool> AddNewValuation(IIndiceValuation indiceValuation)
 
             var attachedValuations = new List<ComponentValuationDao>();
 
-            foreach (var quantity in compositionDao.ComponentQuantities)
+            if (compositionDao.ComponentQuantityDaos != null)
```
monsieurleberre, line 72: When can this be null?
Is this method adding at the same time a composition and a valuation?
I am sure it works, but there are a few code paths that I am not sure are necessary (in line 61, I would exit if no composition is found). Maybe I am missing something

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdaterService.cs
```diff
@@ -34,7 +38,7 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
             {
                 _logger.LogInformation(
                 "Background Service IndiceValuationUpdaterService is running at: {0}",DateTimeOffset.Now);
-                await Task.Delay(IndiceValuationUpdater.PollingInterval, stoppingToken).ConfigureAwait(false);
+                await Task.Delay(_pollingInterval, stoppingToken).ConfigureAwait(false);
```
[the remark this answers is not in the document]
jmc171144: It's to stay coherent with the the polling interval value whenever it's changed during testing (no unit tests, debug testing), but before merging if we set the polling interval to 1 hour may be we can set an interval of 10 or 30 minutes here yes

src/Trakx.Persistence/IndiceDataCreator.cs
jmc171144: It's never null haha it was rest of an old bad implementation
For line 61, I though that could be good to allow to create a new valuation when the composition doesn't exist yet. It's useless for this PR, but I tough that it could be useful to have this option in this function. If you have a problem with that I can remove this part no problem

monsieurleberre: Yes I think it is worth removing it because it will make the code easier to read.
I think it doesn't hurt, for instance when creating a new composition and having to persist its initial valuation, to have 2 functions to call, one to persist the new compo, then one to attach its initial valuation

src/Trakx.MarketData.Server/Data/DataUpdater/IndiceValuationUpdaterService.cs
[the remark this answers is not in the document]
monsieurleberre: cool with me 👍

PR #216, opened 2020-08-24
Asymmetric bearer integration

AymericNoel:


src/Trakx.Common/Utils/MockJwtTokens.cs
```diff
@@ -21,11 +22,12 @@ public static class MockJwtTokens
 
         static MockJwtTokens()
         {
-            var sKey = Encoding.ASCII.GetBytes("ThisIsMySecretKey");
-            SecurityKey = new SymmetricSecurityKey(sKey);
-            SigningCredentials = new SigningCredentials(SecurityKey, SecurityAlgorithms.HmacSha256);
+            var rsa = RSA.Create();
+            rsa.ImportRSAPrivateKey(Convert.FromBase64String(Environment.GetEnvironmentVariable("PRIVATE_BARONG_KEY") ?? "wrong_key"), out _);
```
monsieurleberre, line 26: will you ever need that?
`Convert.FromBase64String(Environment.GetEnvironmentVariable("PRIVATE_BARONG_KEY")`
or should we decide that the key is "wrong_key" and then the public key resulting from it is the one you put in the config later on?

monsieurleberre (approved):
OK for me, I am not sure we need to allow key imports in the mockJwt, am I correct in thinking that if the env variable is set, then the public key in config needs to derive from it?

monsieurleberre (approved):
perfect thanks!

PR #217, opened 2020-08-24
client was calling MockJwt constructor and thrown errors

AymericNoel:
+  MockJwt was called by client and this thrown errors : 

image

Because srcPath doesn't exist on the VM.

+  Server failed because of the `GenerateMockLabels` in the middleware.


monsieurleberre (approved):
I think that will work very well for now, something to keep in mind for readability (in my opinion); you could declare an interface like ITokenGenerator, and have one dev implementation (like MockJwtTokens.GenerateJwtToken) and one for production (which does nothing), then you can inject the ITokenGenerator you want using DI. It will let you test the AuthorisedClient a bit more easily too I think.

PR #219, opened 2020-08-25
Retrieve barong cookie

AymericNoel:
I've tried to inject my new class `CookieRetriever` but I can't because I can't pass the correct parameters to ApiClient via Nswag..


monsieurleberre (approved):
maaaster!!

PR #226, opened 2020-08-27
Issue redeem calculations

monsieurleberre:
A component to calculate quantities involved in issuance and redemption of compositions.
Also removed the `Weight` from `IComponentQuantity`, as it doesn't make sense outside a valuation context.


src/Trakx.Common/Core/ComponentQuantity.cs
```diff
@@ -12,25 +12,23 @@ public class ComponentQuantity : IComponentQuantity
...
                 0, 18);
 
             ComponentDefinition = definition;
             Quantity = unscaledQuantity.ScaleComponentQuantity(ComponentDefinition.Decimals, indiceNaturalUnit);
-            Weight = weight;
 
             Debug.Assert(this.IsValid());
         }
 
+        public ComponentQuantity(IComponentQuantity quantity, decimal? newScaledQuantity = default)
```
AymericNoel, line 26: Is it a constructor that takes a scaled quantity ? It's quite practical

src/Trakx.Drafts/Tools/EthereumBalanceViewer.cs
```diff
@@ -0,0 +1,216 @@
...
+        public EthereumBalanceViewer(ITestOutputHelper output)
+        {
+            _output = output;
+            _flurlClient = new FlurlClient(@"https://api.ethplorer.io");
+        }
+
+        [Fact]
+        public async Task GetEthereumBalances()
+        {
+            var address = "0x9DC57560Dc50cCFB7475B78B594B2444CFB7a1A5";
```
AymericNoel, line 30: Is it the Trakx address for which we're looking for tokens balances ?

AymericNoel (approved):
I've seen that you have added a `IssueRedeemCalculator` but I didn't see where you use it ? 
Is that means that we're going to issue and redeem for real soon ?
The tokens balance retriever is for displaying on the frontend the Trakx balance ?

src/Trakx.Drafts/Tools/EthereumBalanceViewer.cs
monsieurleberre: yes this is the address at which we currently have way too much 😨

src/Trakx.Common/Core/ComponentQuantity.cs
monsieurleberre: yes then you can get an almost clone of the quantity but with changed quantity :)

PR #266, opened 2020-11-06
Modifying way to get version tag for docker packages

AymericNoel:


.github/workflows/dockerpublish.IndexManagerClient.yml
```diff
@@ -41,26 +46,35 @@ jobs:
...
-        uses: actions/upload-artifact@v1
-        with:
-          name: BUILD_NUMBER
-          path: BUILD_NUMBER
+      - name: Get next semver
+        id: bumpSemver
+        uses: anothrNick/github-tag-action@1.26.0
+        env:
+          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          DEFAULT_BUMP: ${{github.event.inputs.semverIncrementLevel}}
```
monsieurleberre, line 54: I think this one is not defined when this is triggered by a pull request => It will be set to minor

monsieurleberre (commented):
1) are you sure we want to run this for each pull request (then at every commit on the PR?), I would have thought, only when merging to dev or master maybe
2) if so that is fine, but it looks like the default bump will be minor on all PRs, I think that is should be patch

No pb otherwise 👍

PR #267, opened 2020-11-08
Add way to get trakx balance and indice supply with binance balance

AymericNoel:
+  I have to do tests on `IEthereumBalanceViewer`
+  In next PR, I will clean IndexManager Client and some functions in controller (like GetTrakxBalance in wrappingController)


src/Trakx.IndiceManager.Server/Managers/EthereumBalanceViewer.cs
```diff
@@ -106,6 +91,7 @@ public async Task<List<AssignedBalance>> GetAssignedBalances(DateTimeOffset navs
                         symbol = balance.symbol,
                         unscaled_balance = descaledAssignedQuantity
                     };
+                    balance.SetBalance();
```
monsieurleberre, line 94: I don't like this honestly, please let me know if there is a very good reason to do it manually.

src/Trakx.IndiceManager.Server/Managers/IBalanceAssigner.cs
```diff
@@ -10,9 +10,21 @@ namespace Trakx.IndiceManager.Server.Managers
     public interface IBalanceAssigner
     {
         /// <summary>
-        /// get the current supply and the Trakx Balance on Ethereum
+        /// Get the current supply and the Trakx Balance retrieved on Ethereum
```
monsieurleberre, line 13: Maybe let's get rid of this entirely, we are not going to be able to get the balances only from ethereum now, we will dig that code from github's history if we need it again.

src/Trakx.IndiceManager.Server/Managers/IBalanceAssigner.cs
```diff
@@ -10,9 +10,21 @@ namespace Trakx.IndiceManager.Server.Managers
...
         /// </summary>
         /// <returns>A list of <see cref="List{IndiceSupplyModel}"/> or null if the different api calls failed.</returns>
-        Task<List<IndiceSupplyModel>?> GetTrakxBalanceWithIndiceSupply();
+        Task<List<IndiceSupplyModel>?> GetTrakxBalanceWithIndiceSupplyOnEtherscan();
+
+        /// <summary>
+        /// Get the current supply and the Trakx Balance retrieved on binance
+        /// </summary>
+        /// <returns>A list of <see cref="List{IndiceSupplyModel}"/> or null if the different api calls failed.</returns>
+        Task<List<IndiceSupplyModel>?> GetTrakxBalanceWithIndiceSupplyOnBinance();
```
monsieurleberre, line 22: So this kind of implies the user has to know to call 
1) get balance from Binance
2) Get balance from Ethereum
3) Get balance from Huobi
...
4) FInd out what balance from which exchange is to be assigned where ...

It would be better to have the balanceRetriever know (hardcoded or configuration) how to return the whole balance at once, or if we have many, to find a way to automatically get it from all of them, then to have one single function to assign the golbal balance given a state of the accounts.

src/Trakx.IndiceManager.Server/Managers/BalanceAssigner.cs
```diff
@@ -29,19 +30,59 @@ public BalanceAssigner(IEthereumBalanceViewer ethereumBalanceViewer, IAdjustment
...
+        {
+            var indiceSupply = await _adjustmentsService.GetIndiceSupply();
+            if (indiceSupply == null)
+                return null;
+            var expectedBalance = indiceSupply.ToDictionary(i => i.Currency, i => i.RealValueInCirculation);
+            List<AssignedBalance>? trakxBalances;
+            try
+            {
+                trakxBalances =
+                    await _ethereumBalanceViewer.GetAssignedBalanceFromBinance(DateTimeOffset.Now.Subtract(TimeSpan.FromHours(2)), expectedBalance);
```
monsieurleberre, line 54: Here, in my opinion, it should simply get all the balances, aggregated and assign them in the best way possible

monsieurleberre (changes requested):
We can have a quick call to see how to try and make the code user friendly, but basically it shouldn't be expected that the user of the the BalanceRetriever has to go throught a list of functions for the different places we can get the balance from. Ideally if we can get rid of some code that we are not going to use that would be nice too.

src/Trakx.IndiceManager.Models/AccountBalanceModel.cs
```diff
@@ -19,11 +19,21 @@ public class AccountBalanceModel
...
+            Source = source;
+        }
         public string Symbol { get; set; }
         public ulong NativeBalance { get; set; }
         public decimal Balance { get; set; }
         public string Name { get; set; }
         public string Address { get; set; }
         public DateTimeOffset LastUpDate { get; set; }
+
+        public decimal UsdcValue { get; set; }
```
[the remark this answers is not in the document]
AymericNoel: These are old properties, I didn't want to delete all of them now, because they are used in old test and in IndexManager.Client (cf  #268)

PR #269, opened 2020-11-09
Add source for nuget packages in Dockerfiles

AymericNoel:
I don't know if we can put in clear the password of trakx bot to import trakx nuget source to resolve problems with github workflow files ?


monsieurleberre:
> I don't know if we can put in clear the password of trakx bot to import trakx nuget source to resolve problems with github workflow files ?

1) we _have to_ put it in clear for the linux based builds, or unfortunately I have not found how else to do it - maybe it is worth trying to build the dockers as part of windows build, rather than linux for that...)
2) they should be provided to the build as secrets cf this in dotnet-core.yml (and removed as part of the build if possible, although that might be superfluous):

```
    - name: Add github nuget source
      run: dotnet nuget add source "https://nuget.pkg.github.com/trakx/index.json" --name "github" --username "trakx-bot" --password ${{secrets.TRAKX_BOT_READONLY_PAT}} --store-password-in-clear-text
    
    - name: Install dependencies
      run: dotnet restore ${{env.SOLUTION_PATH}}

    - name: Remove github source
      run: dotnet nuget remove source "github"
```

PR #298, opened 2020-12-02
Fix : indexmanager go down when 2 users are on same page using Js Interop

### LINKED ISSUE #296: IndexManager client go down when more than one user use it
On stage.trakx.io/indexmanager, when two users are at the same time on the same page when one is connected, the website go down....
image

AymericNoel:
Try to closes #296 

Also Fix the bug when users log in to trakx via indexmanager

See also : https://github.com/trakx/exchange-api-client/pull/6


monsieurleberre (dismissed):

src/Trakx.IndiceManager.Client/Pages/IndexMarketplace/IndexBalances.razor
```diff
@@ -67,4 +71,18 @@
 
     }
     #endregion
+
+    #region exportation
+
+    public async Task CsvExport()
```
monsieurleberre, line 76: definitely!!!

monsieurleberre (approved):
👍

PR #302, opened 2020-12-08
Add authentication on indexmanager with google

AymericNoel:
I have implemented a functionnal way to log-in in client with google, to restrict some pages to non connected user.

Same for server.

All restristed to non trakx.io
But Roles are not including in google response...


src/Trakx.IndiceManager.Client/Pages/IndexComposition/IndexList.razor
```diff
@@ -132,7 +140,7 @@
         {
             IndexModelList = (await ApiClient.GetAllIndicesAsync()).Result;
         }
-        catch
+        catch (Exception e)
```
AymericNoel, line 143:
```suggestion
        catch
```

src/Trakx.IndiceManager.Client/Pages/IndexMarketplace/IndexBalances.razor
```diff
@@ -75,11 +75,11 @@
 
     public async Task CsvExport()
     {
-        await DefaultGrid.CsvExport();
+        await DataGrid.CsvExport();
```
monsieurleberre, line 78: 👍

src/Trakx.IndiceManager.Client/Pages/IndexComposition/IndexList.razor
```diff
@@ -90,7 +91,15 @@
     protected override async Task OnInitializedAsync()
     {
         base.OnInitialized();
-        await GetIndexList();
+        try
+        {
```
monsieurleberre, line 95: strange this thing about the toaster...

src/Trakx.IndiceManager.ApiClient/ServiceConfiguration.cs
```diff
@@ -11,14 +12,15 @@ public static class ServiceConfiguration
...
-            serviceCollection.AddSingleton<IIndiceCreationClient, IndiceCreationClient>(serviceProvider => new IndiceCreationClient(baseUrl, new HttpClient(clientHandler), serviceProvider.GetService<ICookieRetriever>()));
-            serviceCollection.AddSingleton<IIndiceSupplyClient, IndiceSupplyClient>(serviceProvider => new IndiceSupplyClient(baseUrl, new HttpClient(clientHandler), serviceProvider.GetService<ICookieRetriever>()));
-            serviceCollection.AddSingleton<IAddressMappingClient, AddressMappingClient>(serviceProvider => new AddressMappingClient(baseUrl, new HttpClient(clientHandler), serviceProvider.GetService<ICookieRetriever>()));
-            serviceCollection.AddSingleton<IAuthentificationClient, AuthentificationClient>(serviceProvider => new AuthentificationClient(baseUrl, new HttpClient(clientHandler), serviceProvider.GetService<ICookieRetriever>()));
+            serviceCollection.AddSingleton<IIndiceCreationClient, IndiceCreationClient>(serviceProvider => new IndiceCreationClient(baseUrl, new HttpClient(clientHandler), serviceProvider.GetService<IGoogleTokenRetriever>()));
```
monsieurleberre, line 20: Here there should be a common interface to share for authentication (or we can think that we could also have an array of implementations that can get injected, but definitely an overkill for now), both via Google or via Envoy

src/Trakx.IndiceManager.ApiClient/AuthorisedClient.cs
```diff
@@ -25,15 +37,19 @@ internal abstract class AuthorisedClient
...
-            {
-                msg.Headers.Add("User-Agent", CookieRetriever.GetUserAgent());
-                msg.Headers.Add("Cookie", $"_barong_session={CookieRetriever.GetCookie()}");
-                return msg;
-            }
-            var mockJwt = new MockJwtTokens();
-            var token = mockJwt.GenerateJwtToken();
-            
+
+            #region envoy authentication
```
monsieurleberre, line 41: I agree with codacy here, there is too much commented code. It is never simple to do, but Envoy authentication or google authentication are just authentications, and they shouldn't affect the code in too many places, so the comments should naturally go away.

monsieurleberre (changes requested):
Please try to factor authentication and make it easilly interchangeable (as discussed on slack)

src/Trakx.IndiceManager.Client/Pages/IndexComposition/IndexList.razor
AymericNoel: Yes ! try to remove the autorize attribute at the top of the razor page, then request that page without be connected, as the controller has the authorize attiribute, an error will be thrown and then an exception will stop the client...

AymericNoel: Maybe we should initialize components with error toaster in other block than `OnInitializedAsync `...

monsieurleberre: yes that is my feeling too 🤔

monsieurleberre (commented):
I wil try to make it even better tomorrow but that is already a great refactoring 💪👍

monsieurleberre (commented):
There is still a bit of testing to be done here

PR #312, opened 2020-12-17
Update Database Initializer in startup indexmanager

AymericNoel:
Now the database initializer is based on env var `SeedDatabaseOnStartup`  boolean but the `SeedDatabase ` function was inaccesible in production...


monsieurleberre (approved):
OK :)

PR #343, opened 2020-12-21
Add Trakx.Utils package

AymericNoel:
Import of Trakx.utils and replace all occurences in solution.

In next PR, I need to import Trakx.Utils in Trakx.IndexManager.ApiClient ! For the moment I need to clean and import current package in Api repositories


monsieurleberre (approved):
58 files changes was scary, but you made it easy to review, thanks 🐱‍🏍👷‍♂️❤

PR #345, opened 2020-12-23
Patch indexmanager

AymericNoel:
### Now Client is functional and some server routes too

What I've done : 
- Add BasePath injection in Trakx.IndexManager.Api because call were never made because of that
- Add Cookie secure policy in order to get authentication with openId
- Update Trakx.Exchange.ApiClient package version because the old one had errors in `AddTrakxExchangeApi()` with Iconfiguration in parameter


src/Trakx.IndexManager.ApiClient/AddIndexManagerApiClientExtension.cs
```diff
@@ -23,11 +24,13 @@ public static partial class AddIndexManagerApiClientExtension
             var apiConfiguration = configuration.GetSection(nameof(IndexManagerApiConfiguration))
                 .Get<IndexManagerApiConfiguration>();
             Guard.Against.NullOrEmpty(apiConfiguration.BaseUrl, nameof(apiConfiguration.BaseUrl));
-            serviceCollection.AddSingleton<ICredentialsProvider, GoogleCredentialProvider>();
-            
+            var options = Options.Create(apiConfiguration);
```
monsieurleberre, line 27: I think this is the less verbose way to do that:
1) make sure the BaseUrl is decorated with a [Required]
Then:
```suggestion
            services.AddOptions();
            services.Configure<IndexManagerApiConfiguration>(
                Configuration.GetSection(nameof(IndexManagerApiConfiguration)));
```

src/Trakx.IndexManager.ApiClient/AuthorisedClient.cs
```diff
@@ -8,12 +8,16 @@ namespace Trakx.IndexManager.ApiClient
     internal abstract class AuthorisedClient
     {
         private readonly ICredentialsProvider _credentialsProvider;
+        public IndexManagerApiConfiguration Configuration { get; protected set; }
+        private string? _baseUrl;
```
monsieurleberre, line 12: so when is this _baseUrl ever set?

src/Trakx.IndexManager.ApiClient/AuthorisedClient.cs
```diff
@@ -8,12 +8,16 @@ namespace Trakx.IndexManager.ApiClient
     internal abstract class AuthorisedClient
     {
         private readonly ICredentialsProvider _credentialsProvider;
+        public IndexManagerApiConfiguration Configuration { get; protected set; }
```
monsieurleberre, line 11: Can't `IndexManagerApiConfiguration` that be readonly?

src/Trakx.IndexManager.ApiClient/ClientConfigurator.cs
```diff
@@ -0,0 +1,20 @@
...
+using Microsoft.Extensions.Options;
+using Trakx.Utils.Api;
+
+namespace Trakx.IndexManager.ApiClient
+{
+    internal class ClientConfigurator
+    {
+        public ClientConfigurator(IServiceProvider provider)
+        {
+            Configuration = provider.GetService<IOptions<IndexManagerApiConfiguration>>()!.Value;
```
monsieurleberre, line 12: Just fyi, generally injecting the service provider itself is considered an anti-pattern.
I am aware this is done in the Trakx.Exchange.ApiClient, the reason being that the ApiKeyCredentialProvider relies on the PublicClient to get the time on which to base its signature, so we can't inject it into itself...
I dont know if it is possible here, but the preferred way is to inject an `IOptions<IndexManagerApiConfiguration>` and an `ICredentialProvider` (If the whole project needs only one, this should be OK).
Not a big deal 😉

src/Trakx.IndexManager.Client/appsettings.json
```diff
@@ -8,7 +8,7 @@
   },
   "AllowedHosts": "*",
   "IndexManagerApiConfiguration": {
-    "BaseUrl": "https://localhost:44330",
+    "BaseUrl": "https://localhost:44366",
```
monsieurleberre, line 11: This keeps changing if you debug using the IIS server. The other way is to debug the projet itself after configuration the launchsetttings once ad for all.
No big deal either :)

monsieurleberre (approved):

src/Trakx.IndexManager.Client/appsettings.json
AymericNoel: Thanks !

src/Trakx.IndexManager.ApiClient/AddIndexManagerApiClientExtension.cs
```diff
@@ -24,13 +24,14 @@ public static partial class AddIndexManagerApiClientExtension
...
-            serviceCollection.AddSingleton(options);
+            serviceCollection.AddOptions();
+            serviceCollection.Configure<IndexManagerApiConfiguration>(
+                configuration.GetSection(nameof(IndexManagerApiConfiguration)));
 
             serviceCollection.AddClients();
             serviceCollection.AddAuthenticationDependencies(configuration);
 
-            serviceCollection.AddSingleton(s => new ClientConfigurator(s));
+            serviceCollection.AddSingleton(s => new ClientConfigurator(s.GetService<IOptions<IndexManagerApiConfiguration>>()!,s.GetService<ICredentialsProvider>()!));
```
monsieurleberre, line 34: sorry to be a pain: don't do this manually
- IOptions<IndexManagerApiConfiguration> is already registered for resolution by the line 27/28
- register the resolution of ICredentialsProvider by providing the type you want
- register the ClientConfigurator
==> after that you have (given the call to AddClients()) all the dependencies necessary for the injection framework to automatically resolve the clients, client configurator, etc, without having to specify things by explicitly calling the constructor of ClientConfigurator

Ask me on slack if it doesn't make sense ;)

src/Trakx.IndexManager.ApiClient/AuthorisedClient.cs
```diff
@@ -8,16 +8,15 @@ namespace Trakx.IndexManager.ApiClient
     internal abstract class AuthorisedClient
     {
         private readonly ICredentialsProvider _credentialsProvider;
-        public IndexManagerApiConfiguration Configuration { get; protected set; }
-        private string? _baseUrl;
-        protected string BaseUrl => _baseUrl ??= Configuration!.BaseUrl;
+        public readonly IndexManagerApiConfiguration Configuration;
```
monsieurleberre, line 11: clean as a whistle :) 🧹 👍

monsieurleberre (commented):
Almost!!!!!

src/Trakx.IndexManager.ApiClient/AddIndexManagerApiClientExtension.cs
AymericNoel: Yeah of course ! It makes sense and it is logical !

src/Trakx.IndexManager.ApiClient/AddIndexManagerApiClientExtension.cs
```diff
@@ -31,7 +30,7 @@ public static partial class AddIndexManagerApiClientExtension
             serviceCollection.AddClients();
             serviceCollection.AddAuthenticationDependencies(configuration);
 
-            serviceCollection.AddSingleton(s => new ClientConfigurator(s.GetService<IOptions<IndexManagerApiConfiguration>>()!,s.GetService<ICredentialsProvider>()!));
+            serviceCollection.AddSingleton<ClientConfigurator>();
```
monsieurleberre, line 33: Perfect 👍

monsieurleberre (approved):

PR #362, opened 2021-01-04
Add new fields to index and constituent supply models

monsieurleberre:
The aim of this is to get a bit more fields in the Index supply page


src/Trakx.IndexManager.Client/appsettings.json
```diff
@@ -8,13 +8,10 @@
   },
   "AllowedHosts": "*",
   "IndexManagerApiConfiguration": {
-    "BaseUrl": "https://localhost:44330",
-    "AuthenticationMethod": "Google"
+    "BaseUrl": "https://localhost:4001",
```
monsieurleberre, line 11: Can you try to not change that in the future check ins? if you use IIS for testing then yes the port is assigned randomly and you have to change often. If you debug using straight dotnet, then this should be OK

AymericNoel: ok perfect !

PR #365, opened 2021-01-04
Collateral balances

carlos-takeapps:
@monsieurleberre , @AymericNoel I created this draft PR so you can check some changes.

I would like to ask you 2 questions now...

1) Of this PR, most relevant is the second commit. And most important for me, the Price property. 
I was able to make it return a calculated value and all the Tests in the solution still successfully passes. But I had to comment out the line 152 of src\Trakx.IndexManager.Server\Managers\CollateralAssigner.cs. This CollateralAssigner is a little complex so I am not 100% sure what's the impact (again, though all the tests still pass).

2) You mention that you wanted 
  ` decimal? Value => Price * Quantity`
and currently we have 
  ` decimal? Price => (Value.HasValue && Value != default && Balance != default) ? Value.Value / Balance : null; `

Correct me, please, if I am wrong but in this context Quantity and Balance are equivalents (I see in the CollateralAssigner that quantities are assigned to the Balance property). So, if that is the case then the Value property would become
   `decimal? Value => Price * Balance`
and the problem with this is we have now a circular reference Price -> Value -> Price ...

If we convert the Price to a non calculated property, then this would be solved. Is that what you expected?

Maybe I am just getting all wrong so, if that's the case, sorry in advance...


monsieurleberre:
No I think it is correct, I will check tomorrow. But you can see I have a pending PR to disambiguate these Quantity/Balance thing. Also I am finding that assignee too complex and might redo it at some point.

src/Trakx.IndexManager.Server/Managers/CoinbaseBalanceRetriever.cs
```diff
@@ -0,0 +1,53 @@
...
+                var wallets = (await _walletsClient.GetWalletsAsync()).Result;
+
+                wallets.Data.ForEach(w =>
+                {
+                    balances.Add(
+                        new CollateralBalance()
+                        {
+                            Symbol = w.Currency,
+                            Balance = Convert.ToDecimal(w.Balance_whole_units),
+                            Source = w.Name,
```
monsieurleberre, line 39: I would put `CoinbaseCustody` here for the source

src/Trakx.IndexManager.Server/Managers/CollateralBalance.cs
```diff
@@ -8,29 +8,20 @@ namespace Trakx.IndexManager.Server.Managers
...
-        public decimal Value { get; set; }
-        public string QuoteCurrency { get; set; }
-        public string Source { get; set; }
-        public DateTimeOffset LastUpdate { get; set; }
+        public decimal? Value { get; set; }
+        public string QuoteCurrency { get; set; }
+        public string Source { get; set; }
+        public DateTimeOffset LastUpdate { get; set; }
+
+        public decimal? Price => (Value.HasValue && Value != default && Balance != default) ? Value.Value / Balance : null;
```
monsieurleberre, line 24: so as discussed, here I woud rather have the Price Settable and the Value automatic

monsieurleberre (commented):
Yes this is the correct direction 👍

src/Trakx.IndexManager.Server/Managers/CoinbaseBalanceRetriever.cs
carlos-takeapps: OK, sure. Do we have some kind of Enum or list of our sources? I have seen Trakx.Shrimpy.ApiClient.Exchange but nothing similar for Trakx.Coinbase.Custody.ApiClient

monsieurleberre: Not that I am aware, and it looks like when we aggregate, we give a CSV list of all sources involved so that.
I think that is fine, maybe in another PR we can do just that, but I don't think we should complicate that one too much again 👍

carlos-takeapps:
@monsieurleberre I believe that both steps are failing because the Integration Test I added needs 
```
CoinbaseCustodyApiConfiguration__AccessKey
CoinbaseCustodyApiConfiguration__PassPhras
```
as environment vars. 

I believe I can add them to the Azure pipeline. Could you please confirm if I should use the same credentials I am using locally? 
I see the env vars are not present in the dotnet-core.yml file

monsieurleberre:
> @monsieurleberre I believe that both steps are failing because the Integration Test I added needs
> 
> ```
> CoinbaseCustodyApiConfiguration__AccessKey
> CoinbaseCustodyApiConfiguration__PassPhras
> ```
> 
> as environment vars.
> 
> I believe I can add them to the Azure pipeline. Could you please confirm if I should use the same credentials I am using locally?
> I see the env vars are not present in the dotnet-core.yml file

OK I think you are correct. I am a bit against having an integration test at that level, given that we already have some in the coinbase custody client project itself. I would rather see a proper unit test that only fails if the CollateralRetriever itself is not working. If the coinbase custody client is not working, we will see it from the integration test on its repo anyway.

How does that sound?

Other than that, can you please sync the branch to the latest dev before aiming to merge?
Thanks!

carlos-takeapps:
I can't add @AymericNoel as a reviewer.

src/Trakx.IndexManager.Server.Tests/Secrets.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.IO;
+
+namespace Trakx.IndexManager.Server.Tests
+{
```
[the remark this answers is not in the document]
monsieurleberre: > I can't add @AymericNoel as a reviewer.

strange, maybe try adding the dev-tools team?
I changed the permissions to not be personal but team based, I guess this is a consequence

monsieurleberre:
> I can't add @AymericNoel as a reviewer.

strange, maybe try adding the dev-tools team?
I changed the permissions to not be personal but team based, I guess this is a consequence

src/Trakx.IndexManager.Server.Tests/Secrets.cs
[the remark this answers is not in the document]
carlos-takeapps: OK, yes. I added this class in order to run the integration tests. Deleted.

src/Trakx.IndexManager.Server.Tests/Unit/Managers/CoinbaseBalanceRetrieverTest.cs
```diff
@@ -0,0 +1,47 @@
...
+            _walletsClient
+                .GetWalletsAsync()
+                .ThrowsForAnyArgs(new ApiException("API processing error", 500, "Error", default, new Exception()));
+
+            var balances = await _balanceRetriever.GetCollateralBalances();
+            balances.Should().BeNull();
+        }
+
+        [Fact]
+        public async Task GetCollateralBalances_should_return_balance_for_empty_wallet()
```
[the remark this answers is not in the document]
carlos-takeapps: Added more test (to check the number of wallets and the balance) I am considering adding a couple more but I don't want to block you.

src/Trakx.IndexManager.Server/Managers/CoinbaseBalanceRetriever.cs
```diff
@@ -35,7 +35,7 @@ public CoinbaseBalanceRetriever(IWalletsClient walletsClient, ILogger<CoinbaseBa
                         {
                             Symbol = w.Currency,
                             Balance = Convert.ToDecimal(w.Balance_whole_units),
-                            Source = _coinbaseSourceName,
+                            Source = $"_coinbaseSourceName|{w.Id}",
```
monsieurleberre, line 38:
```suggestion
                            Source = $"{_coinbaseSourceName}|{w.Id}",

```

monsieurleberre (dismissed):
Thanks!
You could have added a test of the source name to make sure the code was doing what you expected ;)

PR #367, opened 2021-01-06
Readme env

carlos-takeapps:
final version
synched with dev


src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
```diff
@@ -0,0 +1,154 @@
...
+                        if (!string.IsNullOrEmpty(typename) &&
+                        typename.EndsWith("Configuration"))
+                        {
+                            var varNames = new List<string>();
+
+                            var typeProperties = x.GetProperties();
+
+                            if (typeProperties.Count() > 0)
+                            {
+                                var instance = Activator.CreateInstance(x, new object[] { });
```
monsieurleberre, line 131: Can you explain why you actually create all these objects?

src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
```diff
@@ -0,0 +1,154 @@
...
+
+                                foreach (var property in typeProperties)
+                                {
+                                    var attribute = property.GetCustomAttribute(typeof(ReadmeDocumentAttribute)) as ReadmeDocumentAttribute;
```
monsieurleberre, line 135: I usually prefer this more recent syntax
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/type-testing-and-cast#type-testing-with-pattern-matching
no big deal :)

src/Trakx.Common/Configuration/EnvironmentConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace Trakx.Common.Configuration
+{
+    public class EnvironmentConfiguration
```
monsieurleberre, line 3: Shouldn't this come as a result of fiding a reference to the `Trakx.CryptoCompare.ApiClient` package?
If now have to maintain this file manually I think we lost the whole point of this enhancement?
Sorry, maybe there is something I don't understand about this file...

Also please check if you can remove all the warnings below if this file is needed, thanks

src/Trakx.Common/Configuration/ReadmeDocumentAttribute.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+
+namespace Trakx.Common.Configuration
+{
+    public class ReadmeDocumentAttribute : Attribute
```
monsieurleberre, line 5: No!! You wrote this class
https://github.com/trakx/dotnet-utils/blob/dev/src/Trakx.Utils/Attributes/ReadmeDocumentAttribute.cs
and should re-use it, not duplicate it...

monsieurleberre (changes requested):
I am not sure I understand everything as per my comments, however, you can't start duplicating classes.

src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
carlos-takeapps: Because I passed a value to the attribute so we can have cleaner property names mapped to the env var name

```
        [ReadmeDocument("CryptoCompareApiConfiguration__ApiKey")]
        public string CryptoCompareApiKey { get; set; }
```

So I need to get an instance of the class to get the var name that the property is related to.

I was trying to avoid something like 

```
        public string CryptoCompareApiConfiguration__ApiKey{ get; set; }
```

But I can revert that to this last option, which may be simplier to handle.

src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
carlos-takeapps: OK, yes. I wanted more control about the GetCustomAttribute() returning no values, but I will change and test it.

src/Trakx.Common/Configuration/ReadmeDocumentAttribute.cs
carlos-takeapps: Yes, but I was waiting that package to get published, so in the meanwhile I worked to the local class. I wanted to make sure this approach was approved and have already a new branch for deleting this file and taking the class from the package. I will make that change in this same branch.

src/Trakx.Common/Configuration/EnvironmentConfiguration.cs
carlos-takeapps: I thing initially didn't know -or I missed- about this configuration class, that's why I created a local class to be able to work with.  I think it is also the situation as the following comment, if we get this approach approved, I can go and make the changes in other packages and remove these local files. But I wasn't sure about having so many PRs ongoing on many different branches and get things lost.
For example, I would like to first confirm if it is OK for you to pass the var name to the ReadmeDocumentAttribute (see the first conversation) and then, get into another packages and projects to apply the needed changes. Otherwise every correction on this PR would span updates on all the PRs on different branches being worked. I apologize if I was not clear and tell me, please, if you rather use a different approach.

src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
carlos-takeapps: changed

src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
monsieurleberre: no no, my question is not about adding the name from the attribute, I actually like that 👍. I am wondering why we need to call `Activator.CreateInstance(x, new object[] { })` which seems potentially very expensive.

src/Trakx.Common/Configuration/EnvironmentConfiguration.cs
monsieurleberre: Yes yes for sure, but I saw and approved that PR on dotnet-utils : https://github.com/trakx/dotnet-utils/pull/16
To me it means I am happy with the approach.
When I disagree, I try to explain why, and try to resist the merge, like I am doing now 😉😅

src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
carlos-takeapps: Since the name is paased as a parameter to the attribute, the only way I've found to retrieve the name is by instantiating the class (which forces the call to the attribute constructor) and then retrieve the var name from the object instance attribute. Let me see again if I can use reflection on the type and get the var name without instantiating the object.

src/Trakx.Common/Configuration/EnvironmentConfiguration.cs
carlos-takeapps: OK. I think that first I need to solve the issue with the object instantiation and the var name, once I get that solved, I can adjust the Attribute in the other package, merge and publish, and then reference it from this project and delete the local classes.

carlos-takeapps: Removed

src/Trakx.Common/Configuration/ReadmeDocumentAttribute.cs
carlos-takeapps: Removed

src/Trakx.Common.Tests/Unit/Config/ConfigurationProviderTests.cs
carlos-takeapps: Removed the instantiation

carlos-takeapps:
I thought adding to `Trakx.Common.Tests ` a reference to `Trakx.CriptoCompare.ApiClient` so the `CryptoCompareApiConfiguration ` class is documented on this project's Environment.md . 
The thing is that it would create a circular reference between nuget packages. 
I need to dig a little bit more in detail and I am not sure if we want all the different packages to be referenced by this project in order to document all environment variables. 
Maybe it is better to have here a class with a unit test that may be imported from this package by all other projects so the test runs and create a specific Environment.md for each package.

Let me know your thoughts please.

monsieurleberre:
> I thought adding to `Trakx.Common.Tests ` a reference to `Trakx.CriptoCompare.ApiClient` so the `CryptoCompareApiConfiguration ` class is documented on this project's Environment.md .
> The thing is that it would create a circular reference between nuget packages.
> I need to dig a little bit more in detail and I am not sure if we want all the different packages to be referenced by this project in order to document all environment variables used across all our libraries and packages.
> Maybe it is better to have here a class with a unit test exported from this project, that may be imported from this package by all other projects so the test runs and create a specific Environment.md for each package.
> 
> Let me know your thoughts please.

I feel that this is the reason you created the Trakx.Utils.Testing package.

monsieurleberre:
done separately

PR #377, opened 2021-01-18
Collateral assignment improvements

monsieurleberre:
I still need to build the test checking all the actual quantities, but hope to have corrected a bunch of mistakes


src/Trakx.IndexManager.Server.Tests/Unit/Services/Collateral/CollateralAssignerTests.cs
```diff
@@ -186,7 +191,45 @@ public async Task GetAssignedCollateralBalances_should_return_correct_supply()
             retrievedSupply[0].ConstituentSupplies[0].CollateralQuantity.Should().NotBeNull();
         }
 
+        [Fact]
+        public async Task
```
monsieurleberre, line 195: this test is still being written

PR #383, opened 2021-01-26
Copper Collateral Retriever

### LINKED ISSUE #348: Add Copper.co as an ICollateralBalanceRetriever
Currently we only have Shrimpy registered and working.
This is fine as we only have funds there.
However, this is going to change now that we have been onboarded with Copper.co

I have shared with you a readonly api key in Dashlane.
Please use it to read the balance from copper and add it to the balances we read from shrimpy

AymericNoel:
Close #348


src/Trakx.IndexManager.Server/Data/ServiceConfiguration.cs
```diff
@@ -129,5 +163,24 @@ public static IServiceCollection AddBarongAuthentication(this IServiceCollection
                 });
             return serviceCollection;
         }
+
+        /// <summary>
```
monsieurleberre, line 167: 👍 thanks for all the comments

src/Trakx.IndexManager.Server/Startup.cs
```diff
@@ -50,8 +50,9 @@ public void ConfigureServices(IServiceCollection services)
             services.AddControllers();
             services.AddAllManagerForControllers();
             services.AddDatabaseFunctions();
-
+            services.AddApiClients(Configuration);
```
monsieurleberre, line 53: 👍

monsieurleberre (approved):
Nice one, thanks!

PR #384, opened 2021-01-26
Add Fireblocks Collateral Retriever

### LINKED ISSUE #349: Retrieve Collateral balances from Fireblocks
Currently we only have Shrimpy registered and working as an ICollateraBalanceRetriever.

This is fine as we only have funds there.
However, this is going to change now that we have been onboarded with Fireblocks

I have shared with you an api key in Dashlane for the Api Test worspace.

Please use it to read the balance from Fireblocks sandbox and add it to the balances we read from shrimpy

AymericNoel:
There is only one problem here : Fireblocks returns for each vault a list of Assets and for each asset an id, a balance, etc ... but no symbol.
For example for the moment, we have one vault (name : `default` ) and two assets inside (id : `BTC_TEST` and `ETH_TEST`).
So in the future, we should rename the asset with the symbol of what it contains I think.

Close #349


monsieurleberre (approved):
🕺 yeeepa! 💃

